PR #102, opened 2014-03-20
better way to send chars to putties

ozkanpakdil:
this is better way to send multi command to ssh sessions


kaymer327:
@ozkanpakdil I know this is an old pull request... Also, I'm new to the mRemoteNG code... But I was looking though this to possibly merge into my fork. But I'm not 100% sure about all of the changes.

Your commit in 24f98c883d98bf9d0961afc66b25caa797974821 mentions ctrl c & ctrl d, but with the "current" mRemoteNG, both of those things already work per my testing....

Could you clarify your comment in regards to "this is better way to send multi command to ssh sessions".

Thanks!!!

kaymer327:
OK... So I just found pull request #101 and the "mutli command to ssh" makes more sense now.

kmscode:

In addition, PuTTY may no longer be used in the near future:
http://forum.mremoteng.org/viewtopic.php?f=2&t=3196

lolomin:
Hi,

Sorry to reply to this old and closed thread but what about this feature ?
That's a pity that this feature was developped and then not integrated.
I think mRemoteNG should have this feature like MobaxTerm already have, or integrating ClusterSSH (cssh) same behaviour.
Do you think it could be possible?

Regards,

Laurent

ozkanpakdil:
well they may not merge it to main repo but I already prepared a release with this code and used it around 3 years. you can get it from here https://github.com/ozkanpakdil/mRemoteNG/releases

mrwulf:
I agree- this seems like an obvious and "killer" feature: https://github.com/mRemoteNG/mRemoteNG/issues/611

PR #103, opened 2014-03-22
Multiline userfield and several verifications to avoid exceptions

SagePourpre:
I added the winforms propertyeditorbase from codeproject that allows to easily implement user control or forms for the property grid. 

The real goal was to add mulltiline capability to the user field connection info property so more informations can be stored and easily read for each connections. 

Also, I added some conditions to try and ensure files exists before trying to get versions for putty so it avoid exceptions being thrown.


kmscode:

PR #105, opened 2014-06-01
Made UI more consistent in appearance and follow WinForms design guidelines

aplocher:
Swapped OK and Cancel buttons on the Password and Select Panel forms, so the OK button comes first to match the rest of the app and the established Windows standard.  Also removed flat style controls from Select Panel form, to be consistent.

2014-06-01_03h11_08
2014-06-01_03h05_41


kmscode:
Closing PR. Most changes have been made as part of the c# conversion, others were done in commit: da9680d21027e4b62b1fca9538fd9dff5070d458 

Thanks!

PR #106, opened 2015-03-21
Linux build fixed

krichter722:
Some changes without change of function which allows building in `monodevelop` on Ubuntu 14.10.


kmscode:

The [mono_support branch](https://github.com/mRemoteNG/mRemoteNG/commits/mono_support) builds in Xamarin Stuido, and someone runs via WINE - but is non functional. It does not run directly via mono.

PR #107, opened 2015-06-11
Add vSphere protocol

bontiv:
Add vSphere protocole (need vSphere client installed)
Request MR-274


kmscode:

PR #114, opened 2016-05-31
External Tools Fixes

Brandhor:


sparerd:
Thanks for the PR, Brandhor!

Changes look good - merging. (Our Jenkins instance is having issues checking out some git entries... just a bug of how we had to set it up)

PR #119, opened 2016-06-21
create better Multiple Window Support

tokajer:
panels are good but not very usefull with multiple monitors :)


sparerd:
No new code should be merged into feature/sql branch since it is so old (and not even written in the same language as the current develop branch). If you would like to submit new code, please target the current develop branch.

kmscode:
@tokajer - the diff's are impossible to review. As @sparerd noted, if you re-submit based off of the devlop branch we can review it further. Thanks for the PR!

PR #124, opened 2016-07-08
Users/ay/add editorconfig

alexrster:
Add editorconfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+tab_width = 4
+max_line_length = 180
+charset = utf-8
+end_of_line = crlf
+insert_final_newline = true
+
+[*.{js,json}]
+indent_size = 2
+tab_width = 2


sparerd:
@alexrster Can you explain this PR a bit? Why was it proposed? What does it seek to accomplish?

alexrster:
@sparerd I'm use different indent style settings for VisualStudio (I think I'm not the one and only with such problem) - [EditorConfig](http://editorconfig.org) is the tool (VS extensions) that allow to define these settings on folder level (project/solution in our case). Btw, it is supported by most code editors regardless platform.

kmscode:
@alexrster I see the .editorconfig as a user preference. And since not everyone will have this installed, I've added this to the .gitignore so you can still use it and not worry about checkins: 
https://github.com/mRemoteNG/mRemoteNG/commit/b3bc04251ba98721644ec9634c9e7f9de4b24405

Closing this PR un-merged.

@sparerd - I like the idea though. We were discussing some coding standards. This might be good to adopt. We can always remove the entry from .gitignore if desired.

PR #125, opened 2016-07-08
Add ICA compiler key as solution couldn't be built

alexrster:
- Add ICA compiler key as solution couldn't be built
- Upgrade to .NET 4.6.1


sparerd:
It appears the whole purpose of this PR is to optionally bypass compiling anything to do with ICA?

Did you completely set up your development environment?

https://github.com/mRemoteNG/mRemoteNG/wiki/Development%20Environment

alexrster:
@sparerd nope - I didn't complete the tutorial, sorry... But, you know, ICA is the only thing that prevents solution to be built successfully, out of the box

sparerd:
I agree it isn't ideal, though stripping out an entire protocol from the compiled product doesn't seem make much sense to me either. Unfortunately I'm going to have to decline this pull request

alexrster:
@sparerd I'm ok with your decision, just consider the fact that ICA protocol can be easily enabled for automated builds to be able to ship it (just add ICA to the list of conditional compilation symbols for the build) but disabled on local/dev environment by default to make dev process smooth and decrease  the number of dependencies that needs to be installed on dev box

kmscode:
@alexrster I tried to remove the ICA stuff completely as well, but we do in fact have some users. Since we do have active users, we would likely need to have ICA defined by default (so WE don't mess something up for those users). So the out of the box build would still fail for anyone new coming in to this fresh anyway to look at it... For building purposes, installing the one component and then forgetting about it works well.. :wink: 

There were some other changes in this PR that didn't seem to line up 100% either... Looks like you had some #if RDP definitions and some ICA. And it looks like the Connection.Protocol.ICA.cs was moved/refactored to Protocol/RDP/. Not sure if that was intentional...

There are some other unrelated changes in the PR as well: _process_OutputDataReceived. Feel free to submit another PR if you have other changes that you're working that may be beneficial! Thanks!!

As far as the .NET Framework change... 4.5 is likely the most compatible. We're not using any 4.6.x specific methods at this time and building it with the 4.5 target will let it run on 4.5 and later. A lot of people still don't have 4.6.1 installed so it's an annoyance to the user base to install that prereq when there's nothing that requires it at this point on the code side.

PR #132, opened 2016-07-21
Mr 928 add option to choose engine for encryption

jamesrenaud:
This PR covers two items, [MR-928](https://mremoteng.atlassian.net/browse/MR-928) and [MR-739](https://mremoteng.atlassian.net/browse/MR-739)

I've added to the options panel a Security section, which included moving the option for encrypting the connections file from Advanced into Security. Within the Security section, I also added the option to switch Crypto providers, and created the additional required bits in the settings to support this.

In order to support the AEAD crypto provider, I had to generate a more secure EncryptionKey, but left the old one in place to support backward compatibility. When utilizing AEAD encryption you must use the GeneralAppInfo.StrongEncryptionKey secure string.

Finally I've resolved the creation of an un-encrypted temporary connections file, and the only file that gets written is the already encrypted connection details.


kmscode:
per our conversation, rejecting this PR while you fix the outstanding item. Thanks!

Issue #147, opened 2016-10-05
Possible memory leak: ConnectionTreeModel


sparerd:
There may be a memory leak relating to the disposal of ConnectionTreeModels and their child ConnectionInfo objects.

sparerd:
Removed some bad implementations of IDisposable... those were my fault. Though (un)luckily not the source of memory issues.

sparerd:
Tons of XmlText and XmlAttribute data living on the heap well after the XmlDeserializer has finished its work. This might be the major cause of the leak.

sparerd:
Interestingly, the Xml objects seem to only be left over from initial program load. Need to look at the startup routines to ensure the deserializer is dereferenced.

There is still some area of the app that is holding onto ConnectionInfo objects unnecessarily. Starting the app with a small cons file then opening a very large cons file makes memory spike up, as is expected. However, reloading the smaller cons file leaves memory at the same level as when the large file was loaded.

sparerd:
This may be due to event handlers not being unregistered when the ConnectionTreeModel changes.

sparerd:
More research shows that the event handlers are not likely to be the cause.

sparerd:
It looks like the TreeListView is retaining a reference to the ConnectionInfos that should be disposed.

In a test going from large file loaded to small file loaded, the dominator of ConnectionInfo objects that survived GC is TreeListView (exclusively). Need to figure out why the TreeListView is holding onto references.

Domination path for "mRemoteNG.Container.ContainerInfo" from "Survived objects of the type mRemoteNG.Connection.ConnectionInfo "
Holds 1663352 B in 7846 objects

BrightIdeasSoftware.TreeListView ->
BrightIdeasSoftware.OlvListViewHitTestInfo ->
BrightIdeasSoftware.OLVListItem ->
mRemoteNG.Container.ContainerInfo

sparerd:
The ConfigWindow may also be holding onto ConnectionInfo references after the load until a new item is selected.

sparerd:
Tried calling ClearObjects() on the TreeListView to wipe out the unwanted references. Didn't have the desired effect

sparerd:
Tried calling ClearCachedInfo() on the TreeListView to wipe out the unwanted references. Didn't have the desired effect

sparerd:
Fixed a very small leak regarding property and collection changed event handlers in the ConnectionTreeView.

sparerd:
[...]

PR #149, opened 2016-10-05
Hunting for memory leaks

sparerd:
Connected to #147


sparerd:
It may take a significant amount of time to track down and resolve the main leak. The good news is that the use case for producing the leak is somewhat uncommon and it only lasts for the lifetime of the application (managed heap leak). I'm going to downgrade the priority of #147 and merge what has already been done in this PR.

Issue #168, opened 2016-10-12
confCons.xml connection password is invalid


jumoog:
between 5d964f76383a7a57992a290a9954f24c5b4b66a3 and b934e29dc08b60026a87391c69fc464c1734cdf6 happend a error .. 

the connection password is always invalid even after change it and save the connections

sparerd:
I'm replicating as well. Taking a look. May be connected to #166

sparerd:
Was an issue with how the encryption key was being saved within the application session. resolved now, though more cleanup needs to be done.

PR #169, opened 2016-10-12
ReSharper: code cleanup

jumoog:
also fixing #168


jumoog:
build works

sparerd:
Thanks for submitting this PR!

Would you consider splitting this into multiple smaller PRs? Going through all +7k lines will take us a very long time. At the very least, the fix to #168 must be a separate PR to make it easier to evaluate.

(Don't worry about the Jenkins build failure. Our pipeline is having issues checking out PRs at the moment.)

jumoog:
I'll split PR

kmscode:
@jumoog - Yes! Thanks! 

However, please also note, that we use ReSharper as well. We tend to do smaller changes while we're in the files doing other things.

I have found that sometimes ReSharper will propose fixes and it will cause other problems despite build success and general run-ability of the resulting code. For example, some times adding the "@" to denote a literal string will cause the designer to throw an exception when trying to open that UI. Also, VS will just overwrite many of the changes to the Designer files.

PR #175, opened 2016-10-16
Fix BorderPen color when focused

jvitkauskas:
Previously inactive color was always set, because of missing "else" keyword.

### DIFF
--- a/mRemoteV1/UI/Controls/ListView.cs
+++ b/mRemoteV1/UI/Controls/ListView.cs
@@ -87,8 +87,13 @@ namespace mRemoteNG.UI.Controls
         private void BuildBorderPen()
         {
             if (Focused)
+            {
                 borderPen = new Pen(HighlightBorderColor);
-            borderPen = new Pen(InactiveHighlightBorderColor);
+            }
+            else
+            {
+                borderPen = new Pen(InactiveHighlightBorderColor);
+            }
         }
 
         private void BuildBrushes(DrawListViewItemEventArgs e)


kmscode:
@jvitkauskas This code change shouldn't really do anything... if not focused, then it'll skip to line 91: InactiveHighlightBorderColor.

You code is valid... I'm just not sure what it is intended to fix? Could you please let me know the steps to reproduce and the problem that occurs?

kmscode:
PS - the jenkins build failure occurs because it didn't detect any changes... Likely because:

```  
if (Focused)
  borderPen = new Pen(HighlightBorderColor);
borderPen = new Pen(InactiveHighlightBorderColor);
```  

 AND

``` 
if (Focused)
{
    borderPen = new Pen(HighlightBorderColor);
}
else
{
  borderPen = new Pen(InactiveHighlightBorderColor);
}
```

have the same exact logic/control path.

jvitkauskas:
The line
```cs
borderPen = new Pen(HighlightBorderColor);
```
does not make sense, because it is immediately overwritten by
```cs
borderPen = new Pen(InactiveHighlightBorderColor);
```

See example here: https://repl.it/Dy7a/0

kmscode:
That's what I got for reviewing code while all the kids are still awake and bothering me. I'll merge when I get home.

sparerd:
Taking a look at the ListView class, it doesn't look like it's in use anywhere. We can safely delete the whole class. This was probably left over from before the ObjectListView switch-over.

kmscode:
Fixed in commit: 6e3bfcdf04b8d43f12ca420381b4c9beb059d4ec

Removed unused code.

Thanks!

Issue #186, opened 2016-10-20
Passwords saved with special/international characters will not work


kmscode:
|  |  |
| --: | --- |
| Operating system | Windows 7 x64 |
| mRemoteNG version | 1.74 |

From @amoscatelli on gitter

I am having an issue when using password containing special characters like '£'
if I input them directly it works
if I insert them in the session configuration it won't work and it will say access denied
If I try to login on the same server with a different user with a password without that character it works

kmscode:
From [MR-704](https://mremoteng.atlassian.net/browse/MR-704):
I had some SSH passwords with the symbol ° but they are not passed correctly to the SSH session and I had to type the password manually. 

From [MR-621](https://mremoteng.atlassian.net/browse/MR-621):
saved SSH password containing umlaut will not work
When storing an SSH account's password that contains an umlaut, the automatic login would not work.

It is possible to successfully use it on the interactive login shell that appears after "Access denied".

sparerd:
I can confirm that it is not an issue with serialization or internal storage of the passwords in memory. My guess is the issue comes when attempting to pass the characters to PuTTY.

I wonder if this only occurs when attempting to SSH, or if other protocols have issues.

sparerd:
Live test shows that RDP is also affected

sparerd:
Verified the issue is still present in the current develop branch

sparerd:
RDP is actually not affected. I was using a character that was not supported by windows auth (┬).

SSH is still affected for standard international characters in the current build.

sparerd:
The cmdline args being passed to PuTTY are correct. I'm starting to think this is on PuTTY's side.

sparerd:
Based on packet captures performed during interactive login with putty v.067, I'm not confident that PuTTY is accepting the following characters:  £°úä

What is interesting, is that other users are saying interactive login is working for them, even with these special characters. More research is needed.

sparerd:
Finally got the PuTTYNG project to build. Slogging through the authentication code to figure out of there's some evidence to support my theory

sparerd:
Still not sure why this happens for higher order ascii characters, though I've traced it all the way through to where PuTTY sends the auth data to the remote server. The password was never corrupted at any point along the chain of custody (entry into mRemoteNG -> passing to PuTTY via cmdline -> PuTTY parsing it to send to server).

When capturing packets during session initiation, the following interesting thing happened:

[...]

PR #209, opened 2016-10-28
Investigate password support for international chars

sparerd:
Connected to #186


sparerd:
No "fix" was really implemented as I could not find an issue with mRemoteNG. Created a test to ensure higher order ascii characters are saved correctly

PR #212, opened 2016-10-31
New 1.74 pull from new maintainer 

gentoouser:


kmscode:
@gentoouser - not sure what happened here or to your repo. I've seen it in the past when looking through other forks for other possible fixes/features that might not have been PR'd over. 

We've migrated the entire code base over to C#. So even if you didn't close it out, we would have closed this PR since all that source is still under the old VB code.

If you're interested in contributing in C#, we'd love any further assistance! We don't have much in the way of "formal" contributing guidelines at the moment... But if you keep the PR's small that shouldn't be an issue. If you have something specific that you wanted to work on and want to discuss it first - log in to our Gitter chat: https://gitter.im/mRemoteNG/PublicChat 

Take a look at the open issues, and you'll see some of the direction that we are looking to take (looking to replace MagicDock for example).

Thanks in advance for any future contributions!

Issue #271, opened 2016-11-24
Install package is not using the last installation path


patboy29:
Hi
The install package is not using the last installation path. It want to install in the default directory C:\Program files (x86)\mRemoteNG even if the previous installed version is in another drive or folder.

|Operating system | Windows 10 x64 |
|mRemoteNG version| 1.74 and 1.75 Beta 1  |

sparerd:
Resolved - this will be included in the 1.75 beta 2 release. Please note that it requires registry entries to be present to work. You will not see a difference in functionality when upgrading from <= 1.75 beta 1. When upgrading from 1.75 beta 2 to newer versions, the installer will remember the previous install path.

PR #277, opened 2016-11-29
make installer aware of previous install path

sparerd:
Connected to #271


kmscode:
Maybe a Readme / changelog not is an order for the known issue? I would
think that this will only impact 1.74-1.75b1 -> 1.75 updates since we
uninstall 1.72 before installing any of the new versions...

PR #342, opened 2017-01-11
RDP Optional Minutes to Idle Timeout

countchappy:
Allow users to configure a number of minutes until the session will automatically disconnect them for inactivity.

### DIFF
--- a/mRemoteV1/Config/Serializers/XmlConnectionsDeserializer.cs
+++ b/mRemoteV1/Config/Serializers/XmlConnectionsDeserializer.cs
@@ -487,8 +487,8 @@ namespace mRemoteNG.Config.Serializers
                 {
                     connectionInfo.SoundQuality = (ProtocolRDP.RDPSoundQuality)Tools.MiscTools.StringToEnum(typeof(ProtocolRDP.RDPSoundQuality), Convert.ToString(xmlnode.Attributes["SoundQuality"].Value));
                     connectionInfo.Inheritance.SoundQuality = bool.Parse(xmlnode.Attributes["InheritSoundQuality"].Value);
-                    connectionInfo.RDPMinutesToIdleTimeout = Convert.ToInt32(xmlnode.Attributes["RDPMinutesToIdleTimeout"].Value);
-                    connectionInfo.Inheritance.RDPMinutesToIdleTimeout = bool.Parse(xmlnode.Attributes["InheritRDPMinutesToIdleTimeout"].Value);
+                    connectionInfo.RDPMinutesToIdleTimeout = Convert.ToInt32(xmlnode.Attributes["RDPMinutesToIdleTimeout"].Value ?? "0");
+                    connectionInfo.Inheritance.RDPMinutesToIdleTimeout = bool.Parse(xmlnode.Attributes["InheritRDPMinutesToIdleTimeout"].Value ?? "False");
                 }
             }
             catch (Exception ex)
--- a/mRemoteV1/Config/Serializers/XmlConnectionsDeserializer.cs
+++ b/mRemoteV1/Config/Serializers/XmlConnectionsDeserializer.cs
@@ -487,8 +487,8 @@ namespace mRemoteNG.Config.Serializers
                 {
                     connectionInfo.SoundQuality = (ProtocolRDP.RDPSoundQuality)Tools.MiscTools.StringToEnum(typeof(ProtocolRDP.RDPSoundQuality), Convert.ToString(xmlnode.Attributes["SoundQuality"].Value));
                     connectionInfo.Inheritance.SoundQuality = bool.Parse(xmlnode.Attributes["InheritSoundQuality"].Value);
-                    connectionInfo.RDPMinutesToIdleTimeout = Convert.ToInt32(xmlnode.Attributes["RDPMinutesToIdleTimeout"].Value ?? "0");
-                    connectionInfo.Inheritance.RDPMinutesToIdleTimeout = bool.Parse(xmlnode.Attributes["InheritRDPMinutesToIdleTimeout"].Value ?? "False");
+                    connectionInfo.RDPMinutesToIdleTimeout = Convert.ToInt32(xmlnode.Attributes["RDPMinutesToIdleTimeout"]?.Value ?? "0");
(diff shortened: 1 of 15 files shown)


kmscode:
@sparerd Looks like the confCons.xml related tests are failing... Could you take a quick look at this PR if you have some time? Otherwise, I'll look at it next week most likely...

sparerd:
Yea np - taking a look.

These unit tests seem a bit fragile, wish there was a better way to keep them updated :(

countchappy:
If something needs corrected please let me know :)

mRemoteV1/Config/Serializers/XmlConnectionsDeserializer.cs
```diff
@@ -487,6 +487,8 @@ private ConnectionInfo GetConnectionInfoFromXml(XmlNode xmlnode)
                 {
                     connectionInfo.SoundQuality = (ProtocolRDP.RDPSoundQuality)Tools.MiscTools.StringToEnum(typeof(ProtocolRDP.RDPSoundQuality), Convert.ToString(xmlnode.Attributes["SoundQuality"].Value));
                     connectionInfo.Inheritance.SoundQuality = bool.Parse(xmlnode.Attributes["InheritSoundQuality"].Value);
+                    connectionInfo.RDPMinutesToIdleTimeout = Convert.ToInt32(xmlnode.Attributes["RDPMinutesToIdleTimeout"].Value);
+                    connectionInfo.Inheritance.RDPMinutesToIdleTimeout = bool.Parse(xmlnode.Attributes["InheritRDPMinutesToIdleTimeout"].Value);
```
sparerd: For those that are running the beta, line 490/491 will throw an error due to the XML node not having a value (and .Value will be called on a nullref). Users coming from v2.5 confCons wont have this issue, but it'll be good to keep our beta users happy :)

We should really be adding guards against this everywhere, its just more noticeable in this case.

countchappy: Would a fix for this be checking if the XML nodes are null before referencing `.Value` and if they are, get the defaults?

sparerd: Yea, pretty much. You can even inline this by using the null conditional operator and then the null coalescing operator.

```csharp
collectionOfInts["abc"]?.Value ?? 0
```

countchappy:
Pushed a new commit using the null-coalescing operator. Please keep me posted :)

mRemoteV1/Config/Serializers/XmlConnectionsDeserializer.cs
countchappy: @sparerd I just pushed the commit using the null-coalescing operator. Please keep me posted :)

sparerd: Looking better - still need the null conditional operators. Trying to access the Value property on a null value will still throw an exception. The bonus is that this will fix the failing unit tests.

```csharp
collectionOfInts["abc"]?.Value
```

mRemoteV1/Connection/ConnectionInfoInheritance.cs
```diff
@@ -83,9 +83,14 @@ public bool EverythingInherited
...
 		LocalizedAttributes.LocalizedDisplayNameInheritAttribute("strPropertyNameAuthenticationLevel"), 
 		LocalizedAttributes.LocalizedDescriptionInheritAttribute("strPropertyDescriptionAuthenticationLevel"), 
-		TypeConverter(typeof(MiscTools.YesNoTypeConverter))]public bool RDPAuthenticationLevel {get; set;}
-				
-		[LocalizedAttributes.LocalizedCategory("strCategoryProtocol", 4), 
+		TypeConverter(typeof(MiscTools.YesNoTypeConverter))]public bool RDPAuthenticationLevel {get; set; }
+
+        [LocalizedAttributes.LocalizedCategory("strCategoryProtocol", 4),
+        LocalizedAttributes.LocalizedDisplayNameInheritAttribute("strPropertyNameRDPMinutesToIdleTimeout"),
+        LocalizedAttributes.LocalizedDescriptionInheritAttribute("strPropertyDescriptionRDPMinutesToIdleTimeout"),
```
sparerd: Here's the source of the issue for my last requested change (name/description not showing up in the config window).

You have `strPropertyNameRDPMinutesToIdleTimeout` set for the attribute on line 89, though the actual name for the localized string is `strPropertyNameMinutesToIdleTimeout`. Same for line 90. This will need to be changed to pull in the correct (non-empty) name and description.

mRemoteV1/Config/Serializers/XmlConnectionsDeserializer.cs
countchappy: Oh, I see. Fixing and commiting now.

sparerd: 👍 Looks good! Just fix up the localized name/description fields for the inheritance properties and I think this PR will be good to merge.

mRemoteV1/Connection/ConnectionInfoInheritance.cs
countchappy: @sparerd Hopefully the commit I just pushed fixed that.

sparerd: This section will work now, since the localized string vars were changed.

mRemoteV1/Connection/AbstractConnectionInfoData.cs
```diff
@@ -226,6 +227,15 @@ public ProtocolRDP.AuthenticationLevel RDPAuthenticationLevel
         }
 
         [LocalizedAttributes.LocalizedCategory("strCategoryProtocol", 3),
+            LocalizedAttributes.LocalizedDisplayName("strPropertyNameMinutesToIdleTimeout"),
+            LocalizedAttributes.LocalizedDescription("strPropertyDescriptionMinutesToIdleTimeout")]
```
sparerd: These localizations need to be changed to match the new variable names that were changed in the last commit.
line 230 should use `strPropertyNameRDPMinutesToIdleTimeout`
line 231 should use `strPropertyDescriptionRDPMinutesToIdleTimeout`

Change these two and I think we will be good.

sparerd:
👍  Everything looks good! Accepting this PR

sparerd:
Thanks for your work on this @countchappy! Grats on the 1st PR!

Issue #369, opened 2017-01-25
Reset Layout Option Does Not Reset Notification Pane


jcefoli:
When resetting the layout, it does not clear the notification pane. Should be a workaround to #368, but doesn't resolve that.

**Actual**
1. Reproduce #368
2. Reset Layout ( View --> Reset Layout)

**Expected**
Reset Layout should revert to the default pnlLayout.xml schema that is generated on first startup if it does not exist. If you have the notification pane visible and reset, it is not cleared.

PR #371, opened 2017-01-25
Quick fix for #369 (Reset Layout Ignores Notification Pane)

jcefoli:
Fix for #369 

Reset Layout will now hide the notification pane (if it's opened) and dock it to it's default place (if it's been moved). I've tested this by floating the notification pane, and then resetting the layout. It seems to match its default behavior on initial launch with no preexisting pnlLayout.xml config.


sparerd:
Looks like it works as advertised.

@kmscode any issues with this change?

kmscode:
Nope. Looks good to me. Merging.

PR #372, opened 2017-01-25
Serialize connectioninfo ID

sparerd:
The unique connectioninfo id is now being serialized to the confCons file. This enables us to tie serialized objects to their inflated instances in memory.

This change was made to support the credential manager feature being developed for v1.76.


sparerd:
@kmscode Requesting a quick review from you before merging this into develop for release with 1.75.

Issue #195, opened 2016-10-24
Access to https with self-signed certificates not working


MaRRiK74:
As the title says. Using https links that use selfsigned certs are impossible to open.
The option to 'continue' is missing.

|  |  |
| --: | --- |
| Operating system | Windows 7 x64 |
| mRemoteNG version | 1.75 aplha 3 |

kmscode:
https://mremoteng.atlassian.net/browse/MR-977
https://mremoteng.atlassian.net/browse/MR-995

sparerd:
Need to handle the `NSSError` event thrown by the Gecko browser. Default should be to prompt the user whether they would like to continue.

Not entirely sure how to get the IE browser working yet.

sparerd:
The IE browsers inability to prompt to continue was introduced in commit `a1beff7` that was meant to fix another bug. Trying to find that ticket..

sparerd:
Here is the related bug report for commit `a1beff7`: [MR-706](https://mremoteng.atlassian.net/browse/MR-706)

This may require us to add a connection-specific option for toggling error suppression.

sparerd:
The geckofx documentation isn't that great. Going to need to play around with this for awhile. No guarantees this will make it into 1.75

kmscode:
There is a workaround listed in #259

l0wBoB:
Any news or schedule on this? @kmscode

Platanisiotis:
I'm trying to pitch this software for my company and we're running into this same issue which is a major problem if we were to use it.

kmscode:
From #259:

>Allow embed Firefox(GeckoFx) to access SSL w/invalid certs in Remote Connection HTTPS
>
>Manual fix to use your firefox exception SSL w/invalid certs:
>Copy:
>c:\users%USERNAME%\AppData\Roaming\Mozilla\Firefox\Profiles{RANDOM}\cert_override.txt
>and place it here
>c:\users%USERNAME%\AppData\Local\Geckofx\DefaultProfile\cert_override.txt
>
>Note: https://developer.mozilla.org/en-US/docs/Cert_override.txt

**tentatively scheduled for 1.76**

Alkaceq:
Ïs there any plan to have this fix released?

rscanman:
Hi Guys,

Any chance this can be re-addressed?

JOduMonT:
Funny it is:
[...]

PR #451, opened 2017-03-12
Allow insecure certs for Gecko

kmscode:
Possible fix for #195. At least initial fix. This does work and is straightforward. 

I could potentially put a prompt in the event handler in Connection.Protocol.HTTPBase.cs... But I think that would be massively annoying. 

Check this out to make sure that I didn't mess up any of the config related stuff...

### DIFF
--- a/mRemoteV1/Connection/Protocol/Http/Connection.Protocol.HTTPBase.cs
+++ b/mRemoteV1/Connection/Protocol/Http/Connection.Protocol.HTTPBase.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Windows.Forms;
 using Gecko;
+using Gecko.Events;
 using mRemoteNG.Tools;
 using mRemoteNG.App;
 using TabPage = Crownwood.Magic.Controls.TabPage;
@@ -13,13 +14,14 @@ namespace mRemoteNG.Connection.Protocol.Http
 	{
         #region Private Properties
 		private Control wBrowser;
-		public string httpOrS;
-		public int defaultPort;
+	    protected string httpOrS;
+	    protected int defaultPort;
 		private string tabTitle;
         #endregion
 				
         #region Public Methods
-		public HTTPBase(RenderingEngine RenderingEngine)
+
+	    protected HTTPBase(RenderingEngine RenderingEngine)
 		{
 			try
 			{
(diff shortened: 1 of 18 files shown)


mRemoteV1/Connection/Protocol/Http/Connection.Protocol.HTTPBase.cs
```diff
@@ -178,6 +181,15 @@ private void wBrowser_Navigated(object sender, WebBrowserNavigatedEventArgs e)
 			objWebBrowser.Navigated -= wBrowser_Navigated;
 		}
 
+        //Refernce: https://bitbucket.org/geckofx/geckofx-33.0/issues/90/invalid-security-certificate-error-on
+        private void GeckoBrowser_ValidityOverride(object sender, CertOverrideEventArgs e)
+        {
+            if(InterfaceControl.Info.HttpsTrustInsecureCerts == false) return;
```
kmscode: @sparerd I could likely put a user prompt here if you feel it's absolutely necessary.

mRemoteV1/Connection/Protocol/Http/Connection.Protocol.HTTPBase.cs
```diff
@@ -68,12 +70,13 @@ public override bool Initialize()
...
-                    {
-                        GeckoBrowser.DocumentTitleChanged += geckoBrowser_DocumentTitleChanged;
-                        //GeckoBrowser.Tab.LastTabRemoved += wBrowser_LastTabRemoved;
-                    }
-                }
+				    if (GeckoBrowser == null) return false;
+				    GeckoBrowser.DocumentTitleChanged += geckoBrowser_DocumentTitleChanged;
+				    //GeckoBrowser.Tab.LastTabRemoved += wBrowser_LastTabRemoved;
+
+                    if(InterfaceControl.Info.HttpsTrustInsecureCerts)
```
kmscode: This check might be unnecessary since we're doing it later... But I suppose there is no need for the event handler if we know we're not going to do anything with it.

sparerd:
Awesome! I should have time to review it either tonight or tomorrow

mRemoteV1/Connection/Protocol/Http/Connection.Protocol.HTTPBase.cs
sparerd: We could make the prompt optional. Shown by default, but include an 'always accept certs' sort of check box. Best method for implementing might be to request a class that can handle the override via property injection.

kmscode:
Superseded by #454

Issue #479, opened 2017-03-31
New nodes in connection tree should start in editing mode


pfjason:
When right clicking Connections => New Node or New Folder the new item should start in editing mode so that it can be named immediately without additional clicks.

sparerd:
Note for posterity: this is a regression from previous behavior.

sparerd:
Accepted PR by @pfjason. This will be included in v1.76

nordcore82:
This feature should have been added to 1.76 but is not in the alpha version that was released in dec.

sparerd:
Found the issue - fix should be available in the next v1.76 alpha release

PR #480, opened 2017-03-31
Added a BeginEdit() at the end of the AddNode function.

pfjason:
Resolves issue #479.  When right clicking Connections => New Node or New Folder the new item should start in editing mode so that it can be named immediately without additional clicks.

### DIFF
--- a/mRemoteV1/UI/Controls/ConnectionTree/ConnectionTree.cs
+++ b/mRemoteV1/UI/Controls/ConnectionTree/ConnectionTree.cs
@@ -210,6 +210,7 @@ namespace mRemoteNG.UI.Controls
             Expand(parent);
             SelectObject(newNode, true);
             EnsureModelVisible(newNode);
+            this.SelectedItem.BeginEdit();
         }
 
         public void DuplicateSelectedNode()


sparerd:
Thanks for the PR!

I should have time to review sometime this week.

pfjason:
Take your time, it's one line of code :p

pfjason:
By the way, I use this every day, so I appreciate someone taking over maintenance of the project, it'd been stagnant for a long time.

mRemoteV1/UI/Controls/ConnectionTree/ConnectionTree.cs
```diff
@@ -210,6 +210,7 @@ private void AddNode(ConnectionInfo newNode)
             Expand(parent);
             SelectObject(newNode, true);
             EnsureModelVisible(newNode);
+            this.SelectedItem.BeginEdit();
```
kmscode: "this" qualifier is not needed.

kmscode:
Tested against latest develop. It seems to do the trick, however the edit doesn't work - the UI hangs for a second or two and then we get the following exception:
```
Error saving default connectioninfo property CredentialRecord
   at System.ComponentModel.TypeConverter.GetConvertToException(Object value, Type destinationType)
   at System.ComponentModel.TypeConverter.ConvertTo(ITypeDescriptorContext context, CultureInfo culture, Object value, Type destinationType)
   at mRemoteNG.Credential.CredentialRecordTypeConverter.ConvertTo(ITypeDescriptorContext context, CultureInfo culture, Object value, Type destinationType)
   at System.ComponentModel.TypeConverter.ConvertTo(Object value, Type destinationType)
   at mRemoteNG.Connection.DefaultConnectionInfo.SaveTo[TDestination](TDestination destinationInstance, Func`2 propertyNameMutator)
```

I'm pretty sure @sparerd broke something... 😄

sparerd:
Yep, seems to work ok when targeted on the `master` branch. Something on the `develop` branch is screwing with it.

sparerd:
I think i found what was causing the abnormalities. There was a series of event calls that was causing a high number of host ping checks. I removed the call and the BeginEdit() call seems to be snappy now. Accepting this PR as it currently is.

PR #504, opened 2017-04-12
Add Korean translation

gyuha:
I'm good at using mReamoteNG.
Since there is no translation in Korean, we send you a translation file.

My full name is "Gyuha Shin"

Thanks


sparerd:
Thank Gyuha!

Sorry for the long delay. Accepting this PR now

PR #555, opened 2017-05-15
Update README.MD

polluks:
https://en.wikipedia.org/wiki/Telnet

For the sake of completeness...


sparerd:
Thanks @polluks , merging now

PR #557, opened 2017-05-15
Credential manager upgrade crash

vbfox:
The configuration version was parsed in the credential manager upgrade
process using the current user culture and crashed on cultures where '.'
isn't the decimal separator.


sparerd:
Thanks for the PR @vbfox! merged

Issue #326, opened 2017-01-04
Option to show existing configured connection icon on each tree item instead of pause icon.


johnwc:
It would be nice to visualy see the icon we select in the tree instead of the pause icon for every item.  The play icon is fine for when it is currently connected, or maybe an overlay play icon over the selected icon would be cool too.

sparerd:
Some rewrites of the connection tree have made this a bit easier to implement. Tentatively set for v1.76

fred9176:
+1 :-)

sierramike:
Didn't find this thead before posting #507, so my +1 :)

julianscheideggerleanuxch:
+1 that is the missing feature i looking for :)

kmscode:
This was implemented in 1.76 via #558 
Closing.

PR #558, opened 2017-05-15
Use the connection icon everywhere

vbfox:
**This branch include my fix of #557 as I can't start the application otherwise, i'll rebase as needed depending on what happens to that PR**

I found a little bit hard to find my servers as all connections had the same icon in the treeview so I propose this small change.

* Uses the connection Icon on the tree view & menu
* Overlay a small "Play" icon when connected *(I hesitated with overlaying the same dots as appear in the config but didn't manage to get something looking good, let's say that i'm not an artist...)*

2017-05-15 23_49_20-mremoteng - confcons xml - linux
2017-05-15 23_50_00-mremoteng - confcons xml - linux

Don't know what you think but I feel that it's clearer to have the same icons everywhere.

*(The icon overlay is a variation on the Silk play icon so there should be no license problem as it's already used)*


sparerd:
Looks good to me, thanks for taking the time to implement this @vbfox. You can rebase on develop now since i've accepted #557. 

The only thing I could think to add to this is an option for allowing users to toggle between current functionality and your new setup (with the play icon overlay). If we want to go this route, it might be worth refactoring the ConnectionTree class a bit to allowing plugging in different image list implementations. We could then keep the old implementation as-is and move your implementation to a new class. Let me know what you think.

@kmscode any comments on this PR?

sparerd:
This PR is directly related to #326

kmscode:
@sparerd @vbfox Love it! 

I don't feel the need for an option for the old functionality. The icon setting seemed way under used. This corrects that as far as I'm concerned.

@sparerd - does this still need to be rebased from develop? Looks like there aren't any conflicts... Full disclosure, I only did a brief look into the 2 PRs, but I think this has everything needed.... ?

sparerd:
Yea, the rebase might not be necessary

sparerd:
@kmscode If there are no objections, I'll to the merge today.

vbfox:
Sorry for not responding sooner :)

On the setting I can add it but i'm not sure anyone would want to remove the icon (Maybe i'm biased), still can add if the change is too annoying.

On the refactoring I think i'll try to refactor to get to a point where the icon can be in the file/db.
My target is to do the same as KeePass: A fixed set of icons provided in the app and the user can also import icons that are then stored in the file/db (So they are shared if the file is shared and we can use technology-specific icons for servers)

kmscode:
@sparerd I'm good with the merge.

sparerd:
> I don't feel the need for an option for the old functionality. The icon setting seemed way under used. This corrects that as far as I'm concerned.

> On the setting I can add it but i'm not sure anyone would want to remove the icon (Maybe i'm biased), still can add if the change is too annoying.

I've been surprised in the past about people wanting old functionality. I'm good with accepting the PR as is. If we ever need to re-implement the old style we can always dig through the old code. You are both probably right that no one will complain.

> On the refactoring I think i'll try to refactor to get to a point where the icon can be in the file/db.
My target is to do the same as KeePass: A fixed set of icons provided in the app and the user can also import icons that are then stored in the file/db (So they are shared if the file is shared and we can use technology-specific icons for servers)

@vbfox I like that idea. Would be a good enhancement for the next PR (if you care to do it).

Issue #184, opened 2016-10-19
best option for search


christiandef:
Hi, sorry for my english
it will be better if the search method include search by ip & name host , and highlight the result in the connection panel
I think this would accelerate the search

Best Regards.

Christian Bustos

peterchenadded:
I second this request but also to search the description field.

Looking at the code it seems to be pretty achievable.

peterchenadded:
Will see if i can make the changes myself

kmscode:
@peterchenadded That would be fantastic! Feel free to submit a pull request if you are able to do this:
https://github.com/mRemoteNG/mRemoteNG/blob/develop/.github/CONTRIBUTING.md 

This would be a perfect fit for some other planned changes that we want to include in 1.76 (see #338).

sparerd:
Thanks to @peterchenadded for providing the PR for this issue! This will be available in v1.76

peterchenadded:
No worries,

Is there also any work on a centralised credentials manager which stores
all the credentials and the connections can just reference the data that's
available there?

The folder inheritance based approach at first is a good idea but becomes a
nightmare to maintain, due to the credentials been duplicates for some
folders.

I could a popup like that appearing in visual studio is that for the same
purpose?

Cheers,

On 28 May 2017 12:33 PM, "David Sparer" <notifications@github.com> wrote:

Thanks to @peterchenadded <https://github.com/peterchenadded> for providing
the PR for this issue! This will be available in v1.76

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mRemoteNG/mRemoteNG/issues/184#issuecomment-304488713>,
or mute the thread

sparerd:
Yes, we are working on a credential manager for v1.76. A preview is available already on the develop branch

peterchenadded:
Awesome! looking forward to it. If you need help testing it, am good for it.

Cheers

On 29 May 2017 11:51 PM, "David Sparer" <notifications@github.com> wrote:

Yes, we are working on a credential manager for v1.76. A preview is
available already on the develop branch

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mRemoteNG/mRemoteNG/issues/184#issuecomment-304665911>,
or mute the thread

PR #567, opened 2017-05-24
Added extra logic to search the hostname and description fields of the connection instead of just the name.

peterchenadded:
This pull request is to cover the below feature request:

https://github.com/mRemoteNG/mRemoteNG/issues/184

Minor changes to search the hostname and description field of the connection, instead of just the connection name.

Also added the following three tests:

SearchByDescriptionReturnsAllExpectedMatches
SearchByDescription1ReturnsAllExpectedMatches
SearchByHostnameReturnsAllExpectedMatches

All tests passed after the changes.

Please also note the search performance might be slow due to tree not being indexed but I think it should be sufficient.

### DIFF
--- a/mRemoteNGTests/Tree/NodeSearcherTests.cs
+++ b/mRemoteNGTests/Tree/NodeSearcherTests.cs
@@ -48,7 +48,7 @@ namespace mRemoteNGTests.Tree
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


sparerd:
Thanks for the PR @peterchenadded ! Looks good to me and thanks for writing some unit tests for it as well. Merging

Issue #611, opened 2017-06-26
Multi-exec/Parallel SSH feature


mrwulf:
I would love to see someone implement a way to replicate typing to multiple sessions at the same time. I'm frequently diagnosing issues in small clusters (and love being able to "connect" to a folder and have all contained connections opened) and this would be a killer feature. For an example implementation, check out mobaxterm's multi-execution feature: http://mobaxterm.mobatek.net/features.html

lolomin:
Hi,

I have just found this feature request while searching for the same thing and approve also the development of this feature on mRemoteNG, above all nowadays when clusters are being deployed a lot everywhere.

Regards,

Laurent

larsen0815:
@mrwulf @lolomin You might be interested in https://github.com/jimradford/superputty
I regularly use this to connect to all of my ~40 linux servers and execute commands simultaneously.

sparerd:
This feature will be available in v1.76

PR #644, opened 2017-07-17
Multi ssh

mrwulf:
Relates to issues #611, #101, #102, probably others.
 
Added a toolbar with a textbox to run commands in all open putty sessions. Code is roughly based on excellent work by @ozkanpakdil 

There are bunch of improvements that would greatly improve the user experience (and most are fairly straightforward), but I wanted to get this out there so others can use it too!

Future improvements:
1. Add an option to select/limit connections that receive commands
2. Add an option to always open new connections in their own panel (and hide inner tabs)- this would make docking much simpler and easier to see all sessions at once
3. Be able to move a tab between panels
4. Tool strip button to auto-arrange all panels so they are visible
5. Persist recently used commands (maybe even by panel or connection)
6. Don't require typing to be in the toolbar text box (maybe listen to "selected" putty connection and replicate keystrokes to other connections)


sparerd:
Very cool - thanks for submitting @mrwulf!

I will hopefully have some time to take a deeper look at this tonight and offer comments.

mrwulf:
Awesome, thanks! I haven't done windows development is a while, so please
let me know if I need to make any changes or you have any suggestions.

mRemoteV1/UI/Forms/frmMain.cs
```diff
@@ -961,6 +977,106 @@ private void SetQuickConnectProtocol(string protocol)
 		}
         #endregion
 
+        #region Multi SSH
```
sparerd: Rather than adding this as a region, let's make this its own class. That way we can reuse it elsewhere if needed and frmMain wont get any bigger

mRemoteV1/App/NativeMethods.cs
```diff
@@ -381,6 +384,11 @@ public static int HIWORD(IntPtr value)
         /// <summary>
         /// 
         /// </summary>
+        public const int WM_CHAR = 0x102;
```
sparerd: Include a summary from MS about this message type:

"The WM_CHAR message contains the character code of the key that was pressed. "

mrwulf:
I implemented those changes and also fixed a bug I found- commands wouldn't have been sent to connections established with quick connect (because I was only looking through the ConnectionTree for open connections. Maybe later I'll add another (transient) root node for connections created through quick connect. This hack to tie in connections that are created works well for now.

sparerd:
Looks good now. I'll need to talk with @kmscode quick about which branch to target though. We generally don't push/pr directly into master. We might be able to push this into either a new branch or maybe a patch/smaller deployment.

kmscode:
@mrwulf had some quick conversations with @sparerd. We're likely going to adjust some of our release plans slightly to get this merged in. We'll merge as soon as we can. Thanks for the PR!!!!

kmscode:
@mrwulf discussed this further with @sparerd 

I think there was a lot of users looking for this functionality and the code looks solid enough that we'd like to adjust our release schedule a bit...

Would you mind resubmitting the PR aginst the [patch9 branch](https://github.com/mRemoteNG/mRemoteNG/tree/Patch9)?

Basically patch9 (which was going to be 1.75.7009) will become the new 1.76 and 1.76 will be pushed back to a TBD version number. Patch9 will be going stable "soon".

Thanks again for the awesome PR!

sparerd:
@mrwulf I went through some of the merge conflicts an tried to get things working again. Take a look at the following patch of the proposed changes:


[patch.diff.txt](https://github.com/mRemoteNG/mRemoteNG/files/1456248/patch.diff.txt)

sparerd:
Also, just noticed that the PuTTYNG.exe file is different. What version is it?

sparerd:
Merged this to `develop` in commit c4fdf07

Issue #550, opened 2017-05-10
Double-click rename instead of connecting


Mjolinir:
When double-clicking on an item in the Connections menu, mRemoteNG tries to rename the connection instead of connecting. It does not seem to care if it is a show or fast double-click. This has only been an issue in the past couple versions. 

|||
|--:|---|
|Operating system | Windows 10 x64 |
|mRemoteNG version| 1.75.7000.19194 |

mrwulf:
I'm seeing this too- although only when slowly double-clicking a connection. Since rename is on the context menu and available with F2 it seems like we should probably just disable double-click to rename.

Mjolinir:
I think that would be a good fix

sparerd:
@mrwulf that sounds good to me

sparerd:
Merged fix provided by @mrwulf. This will be available in the next patch

PR #652, opened 2017-07-21
Issue #550 Fix - Don't propagate edit if it didn't come from the menu item

mrwulf:
This prevents double clicks (and anything else but the rename context menu) from trying to edit a tree view item. Effectively stops #550.

This isn't as clean as just setting `LabelEdit = false`, but that would also disable the Rename context menu.


sparerd:
Going to merge this one into `Patch9` as well

Thanks for the PRs!

PR #671, opened 2017-08-05
Theming system revamp

jotatsu:
The idea is to revamp the whole theming system , here are the main changes:

**--On vstheme files--**
- Based on vstheme files, not the best theming format in the world but is what dockpanelsuite is using, it was originally used by a vsix extension in visual studio. Also the theming is kind of independant of the theme format (more ahead)

- Load the vstheme (an xml file with a simple structure theme--Category--Color--background or foreground). The file is query using xpath and store in a dictionary of colors (extendedColorPalette) inside the ThemeInfo Class. Dockpanelsuite palettes are independant of the MremoteNG ones as they use constant style loading. If in the future theme file format is changed or new ones are adopted, an adapter pattern can be used to read to the same dictionary respecting the keys and changing the xpath to w/e is needed. 

- 99% of the colors in the vstheme are not used, but the ones used are probably common to all vstheme files. In the vstheme format is not mandatory to have all the visual studio colors, as internally visual studio merges  the missing colors from the default palette. Same mechanism is used in the MremoteNG system, the base palette is vs2015light. 

**--On develop modifications--**
- Most of the work was done in the UI and Themes folder. Resource/Themes was created with the vsthemes and ColorMapTheme.resx that contains the key-xpath combinations to read and write the vstheme files. 

- DockPanelSuite bumped to the lastest version as i needed some correction to match the foreground color of some controls that library renders. 

- All the "user" themes are based around MremoteNGThemeBase that extends VS2015ThemeBase with some minor adjustments (borders for example). 

- ThemeManager is the main class for theming. Its a singleton for now, maybe in the future use DI? dunno. 

- All themes are in memory all the time, memory impact should be little as they are not the full vstheme xmldoc , but just the Dictionary with the colors needed.

- Some extended controls are glitchy because they are drawn WM paint events, mainly NGComboBox and NGTextBox. If themed NGComboBox should be used in DropDownList mode and NGTextBox sometimes draw in system colors because??? legacy code is messy in this control. 

- I created a copy_themes.ps1 script and added to postbuild_mremotev1.ps1 . I have 0 experience in this script language so i just robocopy the /Resource/Themes folder into the output. 

- The new custom controls should be safe to use in the visual studio graphical ui editor, they look just like the normal ones. 

**--On UI modifications--**
- New custom controls where created for theming. i.e NGButton as a Button, NGLabel as a Label, etc. The new controls should have the same behavior as the original ones, its just that they are redraw to support theming.

- The new controls where created because winforms are not friendly to theme, most of them just offer back and fore color, but not border, highlight, disabled, focused colors. For the most part the theming was done overriding the onPaint and forcing the control to user mode paint. I didnt mess with WM events, those are scary messy legacy code.

- All the list in the app changed from ListView to NGListView, that contol is a child of ObjectListView. This is because ObjectListView is good for theming and ListView is imposible.

- Two main parts are not themed, scroll bars and window title bars. The reason is as far as i know those are mostly WM paint events. There are some options to theme them but seem to be basically hide the system control and do your own from scratch. 

-  Some theming is done on the ApplyTheme methods all over the code. Mainly it has to do with basic forms and panel classes. 

**--On User Experience--**
- By default the theming is disabled on the first launch. That means all the components should draw as if where the original winforms-dockpanelsuite-objectlistview  controls. This is done because there is a lot of magic numbers in the custom controls and i dont know if they will work ok with other UI changes (hdpi, windows high contrast setting, windows scaling, etc ). 

- All the theming is tied to the ThemeManager event ThemeChanged, therefore applying the theme can be done without restart of the app. Exceptions are the dockpanelsuite elements (they require all the dock panels to be closed before changing the theme), some elements that are hard to invalidate on the fly and disable theming. 

- All the embedded dockpanelsuite themes are present: vs2003,2005,2012dark-light-blue,2013dark-light-blue, 2015dark-light-blue  + darcula because i like it, it was taken from https://github.com/strayiker/vs-theme-darcula with some changes. 

- None of the included themes can be modified. User can either drop a new vstheme in the /themes folder or create a new theme based on 2015dark-light-blue or darcula. The new theme can be created, modified or deleted on the options theme page. 

- External Tools was slightly modified with the change to NGListView, it does auto grouping by the executable name. can be disabled deleting ToolsListObjView.AlwaysGroupByColumn = this.FilenameColumnHeader; in ExternalToolsWindow.cs 

- Some Language strings where created in the neutral file but are not created in the other languages. 

Thats all for now, if any question or suggestions i'll try to do or answer when adult life let me :(


kmscode:
@jotatsu I've merged this PR. Thanks!!!!

Looks like there might be a small handful of forms/panels that aren't getting your new theme changes... Specifically (that I've found so far):

- Notifications (which needs to be re-written but that's a different story)
- SSH File Transfer

Would you be willing to contribute further to this fantastic change?

jotatsu:
Hi, i will recheck the windows and tweak some missing things. Me thinks there will be UI bugs after this update, i will tackle those also depending on my off job time. 
Thanks

Issue #536, opened 2017-05-03
mRemoteNG.log file locked


agiz10:
After v 1.74 i finally tried new version. It writes log file regardless if setting "Write log file (mRemoteNG.log)" is ticked or not.
The problem arises when i try to sync config folder between different computers, and with log file locked i get errors as sync program (Syncthing in my case) can not copy it over if program is open on different computers.
This should be a fast fix, but for me, somewhat important one.

|||
|--:|---|
|Operating system | Windows 7 x64 |
|mRemoteNG version| 1.75 aplha 3  |

sparerd:
Thanks for the report - we will take a look

agiz10:
I would just like to do a gentle ***bump*** for this litle bug. Thanks!

kmscode:
Resolved by PR #700

agiz10:
Sweet, thx! :)

PR #689, opened 2017-09-04
Prevented log file creation when writeLogFile option is not set

pedro2555:
Under Options -> Advanced, the option _Write log file (mRemoteNG.log)_ is simply ignored when attempting to create the file.

Setting the log file path to null seems to work as intended. But __it requires a program restart__, maybe a warning for that?

Reported in [issue#536](https://github.com/mRemoteNG/mRemoteNG/issues/536).


pedro2555:
it seems to break every now and then when debugging, I'll continue looking into it.

PR #700, opened 2017-09-06
Prevented log file creation when writeLogFile option is not set

pedro2555:
Under Options -> Advanced, the option Write log file (mRemoteNG.log) is simply ignored when attempting to create the file.

Setting the log file path to null seems to work as intended. But it requires a program restart, maybe a warning for that?

Reported in [issue #536](https://github.com/mRemoteNG/mRemoteNG/issues/536).

### DIFF
--- a/mRemoteV1/App/Logger.cs
+++ b/mRemoteV1/App/Logger.cs
@@ -45,6 +45,9 @@ namespace mRemoteNG.App
 
         private static string BuildLogFilePath()
         {
+            if (!Settings.Default.WriteLogFile)
+                return "";
+
 #if !PORTABLE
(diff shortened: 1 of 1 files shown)


kmscode:
I think this is fine... But I'm going to give it more than a once over when I can to be extra sure. Thanks for the PR!

pedro2555:
Yeah, I think I can see why..

Haven't looked into it, but I think log4net is actually breaking when it tries to log anything, but the exception must be captured by that library (I think it is safe to assume so), and it's transparent to us.

kmscode:
Thanks for the PR!

pedro2555:
@kmscode you're welcome, just remember to close the issue, so the guy that was actually suffering gets the message :).

Issue #487, opened 2017-04-05
Portable version settings not retained


ismailt1:
Settings in the portable version are not retained when opening mRemoteNG on different windows clients.

|||
|--:|---|
|Operating system | Windows 7 x64 |
|mRemoteNG version| v1.75.7007 Hotfix 7  |

TechGuy419:
Same here on Windows 8.1 mRemoteNG Portable 1.75.7003.25226

kmscode:
This is a known issue:
https://mremoteng.atlassian.net/browse/MR-813

We need to ensure that user settings are stored with the exe for the portable edition

Currently settings storage is handled by the framework. For the non-portable edition, settings will be stored in a directory/file similar to this:
C:\Users\[user]\AppData\Local\mRemoteNG\mRemoteNG.exe_Url_e5tysfkxumisr2zfm5hh1bi3b5qn2n4z\[version]\user.config

**This prevents the portable edition from being truly portable.**

ismailt1:
Thanks for the update kmscode. Do you know if we can expect a fix for this in the next hotfix release?

kmscode:
@ismailt1 

No ETA. This requires a fair amount of work. It was attempted a while ago, but caused quite few problems in implementing... 

Reference: https://msdn.microsoft.com/en-us/library/8eyb2ct1(v=vs.110).aspx

kmscode:
This seems to be the best practical implementation that I have found:
http://stackoverflow.com/questions/2265271/custom-path-of-the-user-config

masaeedu:
@kmscode So if I understand correctly, the new location of the `user.config` will not be in the roaming folder, but instead alongside the exe?

kmscode:
@masaeedu 

Please see my comment here:
https://github.com/mRemoteNG/mRemoteNG/issues/346#issuecomment-272504038

Please be sure that scoop doesn't run afoul of GitHubs TOS - I don't want to wake up one morning to find that my repo has been locked for bandwidth abuse.

Since this is likely to be considered a "breaking change", it's unlikely to be done until 1.76 or later.

The goal here to to store the .config file with the .exe for the portable edition. This would allow it to load settings even if loaded off of a USB stick.

masaeedu:
@kmscode I'm reading through that discussion now, and I certainly don't want to cause problems for the project in terms of bandwidth utilization. I'll do some more investigation. 

[...]

PR #701, opened 2017-09-07
Fixes settings save location on portable version

pedro2555:
This is reported on [issue#487](https://github.com/mRemoteNG/mRemoteNG/issues/487)

The portable version doesn't actually save the settings within the application folder, making it not truly portable.

This implements [crdx's `PortableSettingsProvider`](https://github.com/crdx/PortableSettingsProvider), and a custom provider chooser class that basically shits implementation based on the compile time const `PORTABLE` to implement, either `PortableSettingsProvider` class, or the .NET default `LocalFileSettingsProvider` class.

Every setting must then explicitly implement `mRemoteNG.Settings.Providers.ChooseProvider` class in order to benefit from `PortableSettingsProvider`, otherwise they always default to `LocalFileSettingsProvider`.


pedro2555:
Just noticed VS did a mess there with git. There's nothing wrong with the code, its just that VS did a mess there with git (I guess copy paste to/from a linux shell did a mess with the line breaks).

I will leave this here for you to pull and test on your side if you wish. But I'm going to fix that mess in a bit, just so one can read the diff...

PR #702, opened 2017-09-07
fix theme ui

jotatsu:
- Fixed bug in ngcombobox that wasnt rendering DisplayMember elements text in theme mode
- Change in ngbutton icon placement for long text button names
- Verify SSH and notifications file transfer for theme
- Changed ConnectionTree to elipse text based on the width of the panel


kmscode:
Thanks @jotatsu !!!!

Issue #665, opened 2017-07-28
Can not add new connection or new folder


iahim:
|||
|--:|---|
|Operating system | Windows 7 x64 |
|mRemoteNG version| mRemoteNG-Portable-1.75.7008.17935  |

I have just downloaded the portable version and tried to add a new connection or a connection folder. It just doesn´t do anything. 
I tried from the file menu or from the right click menu... and the same issue.

Michaellcosta:
I just installed version 1.75.7008.17907 and am having the same problem.

vordee:
Same problem.

Windows 10 Pro

eli-b:
Same here.
Windows 10 Pro.
Components check shows everything in green except Citrix ICA.
Version 1.75.7008.17907.

bpazzetti:
I have the same isssue, workaround, click on (select) Connections, then you can create a new connection

kmscode:
Resolved via PR #705

PR #705, opened 2017-09-09
Fix proposal for issue 'Can not add new connection or new folder'

pedro2555:
This is reported in [issue#665](https://github.com/mRemoteNG/mRemoteNG/issues/665) and is actually weird just returning, and leaving the user to guess what happened.

My proposal is just to create on the root node if no node was selected; KeePass does the same, I don't find very friendly, but my is its better than doing literally nothing.

There you go, if anyone has a more user friendly solution I'll go for it.


kmscode:
Awesome! Thanks!!!

Issue #207, opened 2016-10-28
Feature request: Specify a target directory for external tools.


MaRRiK74:
Option to specify in which directory an external tool is started, like 'target' in shortcuts:

![image](https://cloud.githubusercontent.com/assets/19833001/19804119/a6b0ea0e-9d0c-11e6-8421-1f883e04ba35.png)

Some tools open other .exe files from within the current directory, at this moment some tools i have installed fail to do that.

|  |  |
| --: | --- |
| Operating system | Windows 7 x64 |
| mRemoteNG version | 1.75 aplha 3 |

sparerd:
This will be available in v1.76

PR #706, opened 2017-09-09
[RFC] Added support for custom working directory in External Tool

pedro2555:
I'm closing #207 and as far as code goes, this is pretty much whats missing; can even be merged straight away with no backward compatibility issues.

But I'm unsure about 2 things:

1. What is `argParser.ParseArguments()` supposed to do? (I've peeked at the code, and it's no easy guess)

2. This, of course, would required adding UI elements to expose this, but just stumbled on [this bit of code](https://github.com/mRemoteNG/mRemoteNG/blob/Patch9/mRemoteV1/UI/Window/ExternalToolsWindow.cs#L196). My intention is to add a new input exactly as there is for _Arguments_, does this require any coordination or how does it work with the internationalisation part? (kinda lost here)
image

### DIFF
--- a/mRemoteV1/Config/Settings/ExternalAppsLoader.cs
+++ b/mRemoteV1/Config/Settings/ExternalAppsLoader.cs
@@ -59,6 +59,10 @@ namespace mRemoteNG.Config.Settings
                     Arguments = xEl.Attributes["Arguments"].Value
                 };
 
+                // check before, since old save files won't have this set
+                if (xEl.HasAttribute("WorkingDir"))
+                    extA.WorkingDir = xEl.Attributes["WorkingDir"].Value;
+
                 if (xEl.HasAttribute("WaitForExit"))
                 {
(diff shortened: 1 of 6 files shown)


pedro2555:
Implemented the whole UI, and the feature itself is ready to ship.

Just missing that internationalisation part I'm not really sure how to go about.

PS: while doing this, ended up finding fixing a small bug where the filename wasn't updated on the ExternalTool instance when changed via the file dialog _Browse_ button.

sparerd:
Retargeted to `develop` branch

sparerd:
@pedro2555 Here is a patch file for proposed changes to solve the merge issues that happened because of the retarget. Let me know if this works for you. Once the merge conflicts are resolved we can start testing and get this merged in

[pr706_patch.diff.txt](https://github.com/mRemoteNG/mRemoteNG/files/1459256/pr706_patch.diff.txt)

sparerd:
Merging this into a custom branch off of `master`, it will still be included in v1.76

sparerd:
This is now merged to the `develop` branch

Issue #197, opened 2016-10-24
latest update removed "Quick Connect" panel for selecting quick connection type


kmscode:
|  |  |
| --: | --- |
| Operating system | Windows 7 x64 |
| mRemoteNG version | 1.73 Beta 2 |

In prior version of mRemote, when typing a host into quick connect and pressing enter, there was a panel at the bottom that allowed you to tab to the correct type of connection - making it very useful to type hostname, enter, tab to SSH2, RDP, VNC etc. 

![image](https://cloud.githubusercontent.com/assets/16654335/19664635/5877ce66-9a0f-11e6-8d12-563745b9065e.png)

sparerd:
I think this one was removed in v1.73 (unreleased) which then got rolled into the v1.74 release when we took over the code base from the previous maintainer. I know a few folks have asked about it.

I have no issue with bringing it back, though I would like to give users the option between the pop-up and drop-down selection styles. Something in the options pages to control it maybe?

kmscode:
This is different. See the quick connect panel at the bottom? That's missing. Basically once you hit enter from the quick connect text box in 1.72, that panel pops up with focus allowing you to tab to the connection type and hit enter again to connect. 

I just copied and pasted the description from JIRA. I'll update accordingly...

welbow:
Would love to see this feature return. It is very handy in heterogeneous environments for those who like to use the keyboard. And thanks to all for taking over the project!

kmscode:
Scheduling for 1.76... @sparerd 's comments are accurate (I thought he was referring to something else initially). 

The overall design of the quick connect toolbar changed where it added a dropdown menu to select the session type. This panel was purposefully removed. Though, I can see how this would be more desirable. 

As @sparerd mentioned, we'll need to make some other design choices / options available to get the best of both worlds -- since the current dropdown menu is a very solid design, but not very intuitive for those that might rely more on keyboard input.

songohan:
I also love to have this feature back. I work in a heterogeneous environment and I can't manage to have all ten thousand devices in my mRemoteNG configuration. So I use quick connect very often and it's a pain to always choose the connection type first (after all, the arrow for the dropdown is very small).
Suggestion 1: add a setting, in which the user specifies if he likes to choose the connection type before (dropdown) or afterwards (panel).
Suggestion 2: just change it back to the panel, because I can’t think of a real use case for the dropdown. If someone really wants to use the mouse, he can also click the connection type in the panel.
Thx

[...]

PR #708, opened 2017-09-10
force call connect on protocol change in quick connect bar

pedro2555:
This pretty much sums up, entirely, what I was proposing in #197.

Basically, as soon as you change protocol, it connects. If, for any reason the address field is not populated, it is simply focused and no connection is initiated; as is the case when clicking the connect button.

### DIFF
--- a/mRemoteV1/UI/Forms/frmMain.cs
+++ b/mRemoteV1/UI/Forms/frmMain.cs
@@ -946,7 +946,10 @@ namespace mRemoteNG.UI.Forms
         private void btnQuickConnect_DropDownItemClicked(object sender, ToolStripItemClickedEventArgs e)
 		{
 			SetQuickConnectProtocol(e.ClickedItem.Text);
-		}
+
+            btnQuickConnect_ButtonClick(this, e);
+
+        }
 		
 		private void SetQuickConnectProtocol(string protocol)
 		{


kmscode:
@sparerd - I'm OK with this in general, but I don't think it should be the default. I think it would be good to have a checkbox (defaulted to checked is OK with me too) option in Tools -> Options -> Connections 

We could change "When closing connections" from radio buttons to a drop down for space savings there.

Thoughts?

sparerd:
I agree with making this a configurable option

pedro2555:
> I agree with making this a configurable option

me too, I'll keep that in mind for when I have a bit

sparerd:
Retargeted to `develop` branch

sparerd:
With the re-target, you will need to abandon the changes made to frmMain and instead add the call to UI\Controls\QuickConnectToolStrip.cs. A lot of ui stuff was moved out of frmMain in the develop branch

kmscode:
@pedro2555... @sparerd and I were discussing this one and reviewing it in relation to the original issue... Took us a little while to come to the conclusion that this is the best solution. :wink:

In that timeframe we had done a lot of refactoring. So I merged your modifications via the commit 75007db6305ce4f0808aa804df1e7c861f03a68f

Thanks again for the PR!

Issue #649, opened 2017-07-20
Cannot integrate any external app


OmisNomis:
I have tried integrating Powershell ISE, Powershell, MinTTY, Git Bash and CMD and none of them are integrating; instead, they open a separate window. 

Have searched extensively and cannot find an answer to this.
|--:|---|
|Operating system | Windows 10 x64 |
|mRemoteNG version| 1.75 aplha 3  |

pedro2555:
First, just be sure here, you are running the external app with the __Try Integrate__ option checked, is that correct?
![image](https://user-images.githubusercontent.com/1645623/30254170-71c6a94e-968c-11e7-9d24-ad1eeafa50be.png)

OmisNomis:
That is correct. It is checked.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

pedro2555:
Ok, thanks.

I asked because you said the apps simply open on a separate window and that is not what I see from here.

Anyway #710 should fix, and a couple other issues with the integration part. If you're able to pull and test on your side it would be great, otherwise the next patch should address this.

kmscode:
@OmisNomis - The only one that I would expect to work based on the current design is Powershell ISE. The others are technically not graphical interfaces and Windows will likely not generate a handle for the UI for us to attach to. However, we are discussing some changes (see #710 if you're interested) that may allow some of them at least to make the attempt. Git Bash doesn't seem to get a MainWindowHandle assigned, so that one will not work even with the proposed changes and other bug fixes. cmd.exe does seem to work after disabling some checks, so "final code" may allow that one to work.

OmisNomis:
Morning! Sorry it has taken me so long to reply - I didn't actually realise there were any follow ups on this issue, I just happened to check just in case.

I have checked and I have version `1.75.7008.17935` which, as far as I can tell, is the latest so should have the fix in #710 ? 

Just in case I am doing something incorrectly; I've documented my steps below...

1) Create a new External tool using Powershell ISE, using `Try To Integrate` set to true. 
2) Create a new connection with the `Protocol` set to `Ext. App` and the `External Tool` set to Powershell ISE
3) Open the connection

When I open the connection this is what I have...

![image](https://user-images.githubusercontent.com/14848394/32094631-59a7914e-baf8-11e7-888a-cd7229e90b37.png)

And Powershell ISE is opened as a separate window. 

Here's my config for reference...

[...]

PR #710, opened 2017-09-11
Fixed integrated cli apps?

pedro2555:
I've noticed #649 myself, when integration cli apps. There was a bunch of swapped names and paths that I fix here.

But, from all I could try [this little change](https://github.com/mRemoteNG/mRemoteNG/compare/Patch9...pedro2555:issue%23649-cannot-integrate-cli-apps?expand=1#diff-c5eb9dabf670787a86a2a706864efecfL66), that seems super dangerous, allows both cli apps to integrate correctly and I could not find an app that could break this.

I've tried blaming that line back in the tree, but got far too deep to continue (6b4f2d3).

### DIFF
--- a/mRemoteV1/Connection/Protocol/IntegratedProgram.cs
+++ b/mRemoteV1/Connection/Protocol/IntegratedProgram.cs
@@ -63,8 +63,9 @@ namespace mRemoteNG.Connection.Protocol
 			    _process.Exited += ProcessExited;
 						
 				_process.Start();
-						
-				var startTicks = Environment.TickCount;
+                _process.WaitForInputIdle(Settings.Default.MaxPuttyWaitTime * 1000);
+
+                var startTicks = Environment.TickCount;
 				while (_handle.ToInt32() == 0 & Environment.TickCount < startTicks + Settings.Default.MaxPuttyWaitTime * 1000)
 				{
 					_process.Refresh();
--- a/mRemoteV1/Connection/Protocol/IntegratedProgram.cs
+++ b/mRemoteV1/Connection/Protocol/IntegratedProgram.cs
@@ -23,7 +23,7 @@ namespace mRemoteNG.Connection.Protocol
 		{
 		    if (InterfaceControl.Info == null) return base.Initialize();
 
-		    _externalTool = Runtime.GetExtAppByName(InterfaceControl.Info.ExtApp);
+		    _externalTool = Runtime.GetExtAppByName(InterfaceControl.Info.Name);
 		    _externalTool.ConnectionInfo = InterfaceControl.Info;
 
 		    return base.Initialize();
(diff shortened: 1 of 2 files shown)


kmscode:
@pedro2555 I'm not comfortable with removing WaitForInputIdle as there could be other apps that this change breaks that we're not aware of. Reference [Process.MainWindowHandle documentation](https://msdn.microsoft.com/en-us/library/system.diagnostics.process.mainwindowhandle(v=vs.110).aspx)

Though, I think we should have our own configurable timeout rather than using Settings.Default.MaxPuttyWaitTime.

I wasn't able to get GitBash to integrate as MainWindowHandle  == 0. cmd throws an exception on WaitForInputIdle:

    WaitForInputIdle failed.  This could be because the process does not have a graphical interface.

I would assume Powershell and MinTTY would have the same problem (either MainWindowHandle == 0 or exception).

It's a bit of a hack, but... I wouldn't be opposed to a check before WaitForInputIdle that looks for known problem processes (like cmd.exe) and skips that check, but performs like a 500ms sleep on the thread before jumping into the while. We should log a message in that instance as well...

What do you think?

kmscode:
Also... I think we should have some other tweaks:

````
if (_handle.ToInt32() == 0)
{
	Thread.Sleep(1);
}
````
Might as well sleep for 1ms as opposed to 0ms... 😕 

Then AFTER the while:

````
if (_handle.ToInt32() == 0)
{
    Runtime.MessageCollector.AddMessage(MessageClass.WarningMsg, "Could not get window handle. Cannot integrate ExternalApp.");
    return false;
}

NativeMethods.SetParent(_handle, InterfaceControl.Handle);
````

pedro2555:
@kmscode Yes, I agree. Removing line 66 is no solution; hence I asked feedback.

I think this needs a bit of rethinking, after reading on the subject for a bit.

Anyway we go, I don't think assuming is ok (in assuming cmd.exe is a CLI app), we could ask, or try, assuming is always dangerous (is known to be the main cause to bugs, btw). Your second approach is a bit more in the likes, but I will look to it more carefully.

For the the other lines, their kinda of broken in the current release. If you want to merge the fix for those, just tell me and I will submit a PR for that.

pedro2555:
@kmscode Ok, lets do like this. I've added the `WaitForInputIdle` line back, and you can just merge the other two typos if you want.

The rest of the weirdos I'll send a separate PR. I've done some trials on this just now, and this requires some sort of _intelligence_ from spawning the process to finding the correct handle.

By the way, in the Git Bash case you mention, the specific issue there is that `git-bash.exe` actually spawns `mintty.exe` and that's the one hosting the terminal window. I was able to search for child processes of `git-bash.exe` and grab the `MainWindowHandle` of `mintty.exe`, and there you go.
image

Anyway, if you wanna merge this (with the `WaitForInputHandle` back), it fixes #649.
And maybe open an issue for an enhancement on the IntegratedProgram class?

Issue #492, opened 2017-04-07
External Tools do get empty %username% %password% fields


seju7:
For my connections I am mainly use the default username and password settings under Tools>Options>Connections>"For empty Username, Password or Domain fields use:".

![grafik](https://cloud.githubusercontent.com/assets/27007110/24797233/9d319318-1b90-11e7-94e2-c630384078ae.png)

When I call an external tools with %hostname%, %username% and %password%, where the config for this connections has empty fields for username & password, empty fields are passed on, instead of the default settings.

|Operating system | Windows 10 x64 |
|mRemoteNG version| 1.75.7000.19194|

sparerd:
I believe it was designed this way (to only use the default user/pass when connecting to the machine), or maybe was just an oversight. I'm going to tag this as a "feature request" since nothing is technically broken, though I agree that we should expect the default credentials to be available everywhere in the application.

robertpopa22:
**Fix available for testing**

A fix for this issue has been implemented in our community fork (`robertpopa22/mRemoteNG`, branch `main`). A beta build including this fix is available from the fork's Releases page.

Please test and report if this resolves your issue.

---

> *We sincerely appreciate the mRemoteNG maintainers and the wonderful community that has kept this project alive and thriving for over a decade. Your work powers thousands of IT professionals worldwide, including us. Our fork builds entirely on your foundation — we're preparing a consolidated Pull Request to contribute these improvements back to the official project. Thank you for everything you've built.*
>
> *— robertpopa22, community contributor*

PR #715, opened 2017-09-14
fallback to empty credentials settings in external apps

pedro2555:
Fixes: #492


pedro2555:
Have no ideia why this is falling. The test suite fails on my side as well.

PR #704, opened 2017-09-09
Fixes settings save location on portable version

pedro2555:
This is reported on [issue#487](https://github.com/mRemoteNG/mRemoteNG/issues/487)

The portable version doesn't actually save the settings within the application folder, making it not truly portable.

This implements [crdx's `PortableSettingsProvider`](https://github.com/crdx/PortableSettingsProvider), and a custom provider chooser class that basically shifts implementation based on the compile time const `PORTABLE` to implement, either `PortableSettingsProvider` class, or the .NET default `LocalFileSettingsProvider` class.

A local Settings class had to be explicitly declared in order to decorate it with the new provider.

From the code I could see, this was attempted in the exact same way before, but with no success. I did not try to fix the existing code because [crdx's `PortableSettingsProvider`](https://github.com/crdx/PortableSettingsProvider) simply works.

PR #720, opened 2017-09-15
First startup ignores cons param

### LINKED ISSUE #676: Portable version startup issue
mRemoteNG supports a command line option to load a specific connections file:

> /cons:PathToConnectionsFile

The problem is that if I use this option on a freshly downloaded portable version of mRemoteNG, it will just open the program but with no connections loaded. If I close it and then open it again with the same command line option, then the connection file does get loaded on startup.

I have narrowed this down to the fact that when you first run the program it creates a new user.config file in
C:\Users\<user_folder>\AppData\Local\mRemoteNG\

When this config file exists, I can load preload connection files just fine, however if I delete the user.config file, then I get the same behavior where no connections get loaded during the first startup.

|||
|--:|---|
|Operating system | Windows 10 x64 |
|mRemoteNG version| 1.75.7008.17935  |

pedro2555:
When launching the application for the first time, the settings actually assume the configured default values, which in turn overrides any setting we set before `Application.Run()`; unless we [force save](https://github.com/mRemoteNG/mRemoteNG/compare/Patch9...pedro2555:676?expand=1#diff-14048bc5d1a12b28dc053ffa640f3f19R314)

Fixes #676

### DIFF
--- a/mRemoteV1/App/Startup.cs
+++ b/mRemoteV1/App/Startup.cs
@@ -218,6 +218,30 @@ namespace mRemoteNG.App
             }
         }
 
+        /// <summary>
+        /// Returns a path to a file connections XML file for a given absolute or relative path
+        /// </summary>
+        /// <param name="ConsParam">The absolute or relative path to the connection XML file</param>
+        /// <returns>string or null</returns>
+        private static string GetCustomConsPath(string ConsParam)
+        {
+            // early exit condition
+            if (string.IsNullOrEmpty(ConsParam))
+                return null;
+
+            // fallback paths
+            if (File.Exists(ConsParam))
+                return ConsParam;
+
+            if (File.Exists(GeneralAppInfo.HomePath + Path.DirectorySeparatorChar + ConsParam))
+                return GeneralAppInfo.HomePath + Path.DirectorySeparatorChar + ConsParam;
+
+            if (File.Exists(ConnectionsFileInfo.DefaultConnectionsPath + Path.DirectorySeparatorChar + ConsParam))
+                return ConnectionsFileInfo.DefaultConnectionsPath + Path.DirectorySeparatorChar + ConsParam;
+
+            // default case
+            return null;
+        }
 
         private static void ParseCommandLineArgs()
         {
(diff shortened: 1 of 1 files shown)


pedro2555:
This probably happens for all params..

mRemoteV1/App/Startup.cs
```diff
@@ -282,29 +309,13 @@ private static void ParseCommandLineArgs()
...
-                    }
-                    else
-                    {
-                        Settings.Default.LoadConsFromCustomLocation = true;
-                        Settings.Default.CustomConsPath = cmd[ConsParam];
-                        return;
-                    }
+                    Settings.Default.LoadConsFromCustomLocation = true;
+                    Settings.Default.Save();
+                    return;
```
pedro2555: @kmscode I'm returning here, before saving the settings file, simply because this was done as such in [`Startup.cs`](https://github.com/mRemoteNG/mRemoteNG/blob/d15e444cb79aeb8f6e0b90b5687276e71ddc1400/mRemoteV1/App/Startup.cs) lines [293](https://github.com/mRemoteNG/mRemoteNG/blob/d15e444cb79aeb8f6e0b90b5687276e71ddc1400/mRemoteV1/App/Startup.cs#L293), [299](https://github.com/mRemoteNG/mRemoteNG/blob/d15e444cb79aeb8f6e0b90b5687276e71ddc1400/mRemoteV1/App/Startup.cs#L299), and [306](https://github.com/mRemoteNG/mRemoteNG/blob/d15e444cb79aeb8f6e0b90b5687276e71ddc1400/mRemoteV1/App/Startup.cs#L306) as of d15e444cb79aeb8f6e0b90b5687276e71ddc1400, which is where I'm branching from.

But I don't think that is correct, there is a bunch of stuff that can be set by the console parameters, and since we return here, we're basically ignoring them. Is this intended?

kmscode: I agree. That return should be removed. 

Fix that and I'll accept the PR.

Thanks!

pedro2555: Will do, expect that in a bit.

pedro2555:
Should be ready now, fixes #676 on my side.

And should have no problem merging with #704.

kmscode:
I need to look at/test #704 a bit more.

pedro2555:
@kmscode ok, hit me up if you have in doubts. It is actually pretty simple.

Issue #482, opened 2017-04-03
Folders do not pass through credentials to connections


JangoBritt:
So folders in the Connections list have the option to configure credentials, but when you create a connection under that folder, it doesn't pass through the credential from the folder... is this a bug?

sparerd:
Just to confirm, is user/password/domain inheritance set on the connection? Also, which version of mRemoteNG are you running?

TechGuy419:
I am seeing this as well but only on the General Root config.  The subfolders pass the credentials just fine with inheritance enabled.  It appears that after I exit mRemote and launch it again that there are extra characters added to the password.

Windows 8.1 
mRemoteNG Portable 1.75.7003.25226

sparerd:
I am seeing a bug with the display of inheritance options for folders/connections that are direct children of the root node. These should not have inheritance as an option.

Example:
- Connections (the root node)
   - Connection1
   - Folder1
       - Connection2

We should not be able to set inheritance on `Connection1` and `Folder1` - the button should be greyed out and the inheritance should be completely off. Currently, we can set inheritance. I'd consider this a bug.

> It appears that after I exit mRemote and launch it again that there are extra characters added to the password.

@TechGuy419  Not able to replicate this at the moment. Are you using the SQL feature, or saving to an xml file?

TechGuy419:
I am saving to an xml file.  Below is a portion of the user.config located in AppData.  I see that the value is encrypted but I feel that this is what is causing the issue.

<setting name="ConDefaultPassword" serializeAs="String">
                <value>723mA4QCB9Xofk7O9v+Zy2mxfu2kAq0MZ02ubsVX6P4=</value>
            </setting>

Your thoughts?

TechGuy419:
I was setting this on Connections (root).  This was so when I type a system name in the quick connect it would us the default credentials instead of prompting me.  I do not add every system I connect to daily to a connection folder.

- Connections (root) - (setting specific UN,PW,DN for quick connections)
     - SomeFolder (setting specific UN, PW, DN)
          - Connections (inheriting all from SomeFolder) 

The issue as only with the password being set on the the Connections (root).  The other password entry on SomeFolder was retained in the portable connection file and has been working as expected.

sparerd:
Ahh I think I see now. You set the user/pass/domain settings on the default connection info (highlighted in orange):
[...]

PR #724, opened 2017-09-17
Default connection password must be decrypted before first use

sparerd:
Potential resolution to one of the problems brought up in #482 

Rearranged some program initialization code to ensure the ConDefaultPassword property in the Settings class is decrypted before it is assigned to the singleton DefaultConnectionInfo which is used through the app.

### DIFF
--- a/mRemoteV1/App/Startup.cs
+++ b/mRemoteV1/App/Startup.cs
@@ -42,8 +42,6 @@ namespace mRemoteNG.App
             ParseCommandLineArgs();
             IeBrowserEmulation.Register();
             GetConnectionIcons();
-            DefaultConnectionInfo.Instance.LoadFrom(Settings.Default, a=>"ConDefault"+a);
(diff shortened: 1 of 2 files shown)


pedro2555:
I guess the issue was not having the encryption key, at the time of decryption.

Since you appear to have moved the same code to latter in the logic. Am I correct?

sparerd:
Sort of -

The DefaultConnectionInfo object was being created before the SettingsLoader ran. The SettingsLoader is responsible for decrypting the ConDefaultPassword. Since these were out of order, the DefaultConnectionInfo was getting the encrypted string and then the SettingsLoader was decrypting the string saved in the Settings class. Since the DefaultConnectionInfo was already created by this point, it never received the decrypted string.

pedro2555:
Yep. Thats exactly it.

kmscode:
@sparerd Changes are simple enough, I think they are good to merge. 

I'll open a new issue for 1.76 to track implementing SecureString and only decrypt on use.

sparerd:
Sounds good to me - merging

PR #728, opened 2017-09-22
Fixed #529

### LINKED ISSUE #529: Erratic Tree Selection when using SQL Database
When using the configuration connected to SQL Server, the Tree selection on the left is changing very often, even if i dont click on another tree item.

It seems there is an autorefresh which does not occurs when using the xml file mode.
This screen refresh is a problem because when editing a server or a connection info, after the refresh, nothing is selected in the tree, so the configuration window is not in sync with the tree, so it makes the save item to fail (no error, but nothing saved, or in a place i do not know).

|||
|--:|---|
|Operating system | Windows 10 pro x64 |
|mRemoteNG version| Prerelease v1.75.7005 (Hotifx 5)  |

dekelMP:
Made Runtime.LastSqlUpdate public and set the connectionTree update to happen only if the user is not the last updater


sparerd:
Was finally able to test this today, looks good! Merging

Issue #663, opened 2017-07-28
Bug on connections panel with first putty session


cryptage21:
Hi,

I've detected a bug with SSH sessions (Putty).
After launching mRemoteNG, if you start a SSH connection, it's working perfectly.

To open a second connection, I click on the "connections panel" and when I want to write the name of the session, it's writing in the opened putty tab instead of selecting the name of the session in the list.

After another click on the list, that's good. With 2 or more opened tabs, it's working correctly too.
The problem is only when there's only one putty session.

Could you please check this issue ? I'm using mRemoteNG 1.75 Hotfix 8.

Best regards

kmscode:
This is just another scenario caused by focus issues with puttyng. See #290

Issue #603, opened 2017-06-20
Importing RDG files from RDCP 2.7 to mRemote 1.75


SAMIT2016:
I need to import .rdg file and i have tried importing it and also opening it and it fails both times. With import the error is - cannot import the file. With opening it the error is - you are trying to load a connection file that was created using an very early version of mRemote, this could result in a runtime error. if you run into such an error, please create a new connection file!

can someone help?

|||
|--:|---|
|Operating system | Windows 7 x64 |
|mRemoteNG version| 1.75 aplha 3  |

kmscode:
@SAMIT2016 Do you mean Remote Desktop Connection Manager?

This is likely a duplicate of #225

Issue #225, opened 2016-11-07
Can not import RDCMan v2.7 build 1406.0 rdg file


deathrush:
|||

2016-11-07 18:00:51,144 [1] ERROR- App.Import.ImportFromFile() failed:1
Unsupported schema version (3).
2016-11-07 18:00:51,150 [1] ERROR- App.Import.ImportFromFile() failed:2
Unable to cast object of type 'mRemoteNG.Tree.Root.RootNodeInfo' to type 'mRemoteNG.Container.ContainerInfo'.

|--:|---|
|Operating system | Windows 10 x64 |
|mRemoteNG version| 1.74  |

sbgillett:
Is this not a bug?

sparerd:
mRemoteNG only ever had support for importing RDCMan v2.2 files. Supporting v2.7 imports would be a new feature.

MoeSaidi:
+1 for this; would be nice to be able to easily migrate from RDC Manager to mRemoteNG

ThoHah:
What is the status here?
This was a request in 2016...

sparerd:
This will be included in the next major release, v1.76

PR #754, opened 2017-10-28
MR-225 Can not import RDCMan v2.7

brucetp:
Referres to [#225](https://github.com/mRemoteNG/mRemoteNG/issues/225) & [#663](https://github.com/mRemoteNG/mRemoteNG/issues/603) 
I've modified the file import to be able to import from RDCMan v2.7 (while keeping the v2.2 in tact).   I also created all the test cases for v2.7.  I did modify the test_RDCMan_v2_7_schema3.rdg embedded resource because the properties in that file didn't match the v2.2 resource and thought the tests should be equivalent.   All tests pass.  Give it a try and let me know.  Thanks


brucetp:
Well, I think the only reason this is failing is because I added a new file to the Test Project.   Let me know if there's something I need to do.

sparerd:
This is weird - I don't know why it still thinks there is a merge conflict

sparerd:
Ok, should be good now. I don't know why it still thought there was a merge conflict

mpriesol:
Hello, thanks for this enhancement. 
I've just installed 1.76.0.25712 and tried import rdg from 2.7, but no success.
So I was testing a bit and found out, that the import doesn't support logon credentials saved to file. In other words, RDC manager offers an option to use saved profiles as a logon credentials. If you use these, it creates following entry in rdg file:
```
      <logonCredentials inherit="None">
        <profileName scope="File">domain\username</profileName>
      </logonCredentials>
```
If you use local credentials instead, additional tags are present, like password and domain.
```
        <logonCredentials inherit="None">
          <profileName scope="Local">Custom</profileName>
          <userName>myusername</userName>
          <password />
          <domain>mydomain</domain>
        </logonCredentials>
```
Technically, if mRemoteNG`s Import function doesn't find <domain> inside <logonCredentials> tag, it fails.
If domain tag is present, it will import data.

anhlqn:
I deleted all the profiles in profile management and was able to import into mRemoteNG

TiloGit:
can't import from 2.7 as well. try to modified the .rdg file but not luck. 
https://www.reddit.com/r/mRemoteNG/comments/9x1fm2/rdcman_27_file_import_fails/

> 2019-04-04 16:38:16,461 [1] ERROR - Error occurred while importing file 'C:\Users\tilos\Documents\TiloRDPs_TiloForImport.rdg'.
> Exception has been thrown by the target of an invocation.
> 2019-04-04 16:39:32,715 [1] ERROR - Error occurred while importing file 'C:\Users\tilos\Documents\TiloRDPs_w_domain.rdg'.
> Exception has been thrown by the target of an invocation.

TiloGit:
I could import my 2.7 file after replace: 

```xml
<logonCredentials.*?</logonCredentials>
>>with (via NP++)
<logonCredentials> </logonCredentials>
```

hubertsvk:
hi,
is the problem resolved? I have RDCman 2.9 and cant import.

PR #760, opened 2017-10-30
MR-747 putty sessions.watcher.start watching failed. not found

brucetp:
The issue was that the FileSystemWatcher for the Putty Sessions (non-portable) was being initiated on a path that didn't exist, so an exception was being thrown.  I added a check to determine if the path exists before the FileSystemWatcher is initiated.

The other issue is that the full path of putty.exe was being combined with the Putty.conf and created an incorrect path.  (var puttyConfPath = "path_to_exe\PuTTYNG.exe\putty.conf").   I modified the statement to only combine the folder of the exe to the conf file.


kmscode:
Awesome! Thanks for the PR! Looks good. Merged!

Issue #152, opened 2016-10-07
Add to External Tools Toolbar option


kmscode:
Received from user via e-mail...

Currently when one enables “External Tools Toolbar” under View, all of the external tools show up on the toolbar. I would like to be able to select which external tools will be available on the toolbar.

My suggestion is to create a new toolbar where tools can be added to it, by selecting a check box called something like “Add to External Tools Toolbar” under External Tools Properties. I have made the following mockup form screen shots to hopefully better illustrate my idea. I know that the example below is not the best but I hope it helps…

![image](https://cloud.githubusercontent.com/assets/16654335/19196113/3c05c4f8-8c82-11e6-921b-053468154f56.png)

I still want the External Tools to be available as they are, when right clicking on a connection and being able to run the commands against them.

sparerd:
This will be available in v1.76

PR #775, opened 2017-11-06
MR 152 add to external tools toolbar option

brucetp:
Refers to #152 
Added the "Show on Toolbar" option to the External Tools Dialog box.
Added the property to the ExternalTools object and defaulted the property to true for new tools. 
Added events to show the the changes immediately upon clicking the checkbox.
Added required code to support multiple languages BUT I did not update each language resource file..  unfortunately I'm not multilingual  :-(

### DIFF
--- a/mRemoteV1/Config/Settings/ExternalAppsLoader.cs
+++ b/mRemoteV1/Config/Settings/ExternalAppsLoader.cs
@@ -82,6 +82,11 @@ namespace mRemoteNG.Config.Settings
                     extA.TryIntegrate = bool.Parse(xEl.Attributes["TryToIntegrate"].Value);
                 }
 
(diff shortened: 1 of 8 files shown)


brucetp:
jenikins is giving me an error...   but I'm not sure why
\packages\NUnit.ConsoleRunner.3.6.1\tools\nunit3-console.exe' cannot be found

sparerd:
Thanks for the PR Bruce - taking a look at Jenkins

sparerd:
It was a stale tool path in the jenkins build file. Since longer lived branches (develop) still had the old nuget packages cached, they were succeeding. New branches didn't have the versions of nunit that the build file wanted. Should be all good now, but I'm going to need to find a nicer way to keep the build file updated.

Issue #784, opened 2017-11-08
Placement of GUI Elements interferes with translated texts length


farosch:
Some GUI Elements on the Settings-Pages **Notifications**, **Connections** and **Update** are placed in a way, that they overlay translated texts. The text fields marked red will collide with other elements if these are too long.
![conn](https://user-images.githubusercontent.com/22905106/32577348-f124cfc4-c4d9-11e7-8c3e-8e6c304e9cd5.png)
![not](https://user-images.githubusercontent.com/22905106/32577349-f142d8ac-c4d9-11e7-9515-db7a34b1c4c9.png)
![upd](https://user-images.githubusercontent.com/22905106/32577350-f15d1794-c4d9-11e7-8756-0d02f8d1a932.png)

These should be rearranged to something like this:
![connfix](https://user-images.githubusercontent.com/22905106/32577361-fd1661b2-c4d9-11e7-806e-0ed52174fed0.png)
![notfix](https://user-images.githubusercontent.com/22905106/32577362-fd34cff8-c4d9-11e7-8239-794ad0ea3d99.png)
![updfix](https://user-images.githubusercontent.com/22905106/32577363-fd530522-c4d9-11e7-8810-bc748080d93f.png)

kmscode:
Did you make the changes to move the elements already? If so please submit a pull request! Looks like it should be straight forward and easy to accept/merge.

If not... I agree with your design suggestions and will look to include them soon.

sparerd:
Will be included in v1.76

PR #791, opened 2017-11-09
Rearranged settings elements

farosch:
Rearranged some elements from the settings to prevent that translated strings overlap with these elements. Issue #784


sparerd:
Thanks for resubmitting @farosch

PR #796, opened 2017-11-10
Further design optimizations on settings

farosch:
Small layout fixes to frmOptions and option pages:
_-unified button heights
-fixed tab order on some option pages elements
-resized "Launch PuTTY"-button under  Options>Advanced to display PuTTY Icon in full size
-resized main panel in frmOption to fit size of option panes
-reduced size of bottom panel in frmOptions to waste less space
-fixed distance to form border of"OK" and "Cancel" buttons in frmOptions
-unified all border styles of panels in frmOptions to none. (This makes the design look a little bit more flat and cleaner, also removes the strange edge  in the title bar where the two panels meet)
-set width of deviders on frmOptions to 1px (also looks a lot cleaner)
-fixed alignment of some elements in options_

Some of these changes can be seen in the screenshots:
**Before:**

**After:**


sparerd:
Looks good, thanks @farosch !

Issue #182, opened 2016-10-19
Rewrite database (SQL) connection storage


kmscode:
Feature was experimental to being with. Lots of bugs/feature requests in JIRA (see list below).

In addition, would be beneficial to have support for MySQL/MariaDB, and/or SQLite. More general DB handling code could be implemented in order to support multiple DB types.

JIRA tickets:
https://mremoteng.atlassian.net/browse/MR-923
https://mremoteng.atlassian.net/browse/MR-923
https://mremoteng.atlassian.net/browse/MR-923
https://mremoteng.atlassian.net/browse/MR-923
https://mremoteng.atlassian.net/browse/MR-923
https://mremoteng.atlassian.net/browse/MR-923
https://mremoteng.atlassian.net/browse/MR-923
https://mremoteng.atlassian.net/browse/MR-923
https://mremoteng.atlassian.net/browse/MR-923

sparerd:
todo: investigate using Entity Framework to handle DB persistence

ysqi:
my store, thank you a  lot.

1. Can refresh config from DB by user , not restart NG.
2. Design a feasible solution for sharing configuration
3. #336 
4. Can define tab backupcolor

mrPsycho:
+1 for different than MS SQL engines. 
would be great to add mysql-compatible, or postgres. 
even firebird.

mbirth:
+1 for SQLite support (on shared storage)

jotatsu:
Candidate for 1.8 as an improvment of 1.77 credential repository

merarischroeder:
Google Drive support would make sense too. You login with your company's credentials. The storage could either be separate .config files in a folder, or they could be Rows in a Spreadsheet. The good thing with individual .config files is that it's easy for Managers to control access to credentials.

merarischroeder:
I think an architecture that standardises and centralises on SQLite would have a lot of benefits.

![Data Architecture drawio](https://user-images.githubusercontent.com/1159081/141665531-e8c73f62-ee4a-47da-8db7-95579e2105ff.png)

Details:

- The Application only connects to the SQLite database. With WAL mode, and encryption.
- Some Providers can be either built into the Application
- An import/export provider is simple and enables what the application can already do today
- A sync provider work with a central repository to distribute configurations. This runs in the background.
- Providers can also be made externally to mRemoteNG. As long as the password is known (or no password is set), its simple to create Import/Export programs and Sync programs.
[...]

PR #799, opened 2017-11-12
Implemented option for saving connections on edit

### LINKED ISSUE #340: Save Configurations without Restarting
Operating System Windows 10 x64
mRemoteNG Version 1.72

It would be really great if settings for new connections were saved immediately after making them.  I create a lot of connections and sometimes the program crashes and I will lose a bunch of connections I've made.

Steps to Reproduce:
1 - open mRemoteNG
2 - create a new ssh connection, add hostname username/password and port
3 - open connection to see its working
4 - crash the program
5 - reopen and notice the new connection and other settings have reverted to the state before any new changes were made.

I don't remember if this was done before I don't remember having this issue in the past, I have used this program for several years now.

Other than that wonderful program thank you very much for your efforts.  I would also be happy to help with the issue if that is of interest.

sparerd:
Gives the user an option of whether to save the currently loaded connections when an edit is made (connection record added, deleted, or updated). This contains some significant refactoring to make the connection load/saving less confusing and more centralized.

Resolves #340


kmscode:
Rather then an option, this should be the only way we save now. We should also remove the "auto save every x minutes" option.

sparerd:
I've been mulling it over and I'm not sure about removing the other save options. However, I can't think of a specific reason that I feel uneasy about it. I'd be ok with making save-on-edit default to enabled (default is disabled at this point in the PR). It might even make sense to make it an either-or choice between save-on-edit and other options (save every X minutes)

kmscode:
I completely understand... I had that feeling as I was typing the comment. I think it's basically because the current save on exit or save at interval have been there forever. But... The default save on exit causes issues from time to time with zero'd out confCons - and save at interval is a band-aid for that. 

I think with this PR you've solved a long annoying problem and we should just move forward.

However, if you still feel uneasy:
Make save on edit the new default.
Save every X could still be available. If this is enabled, then save on exit still needs to be enabled as well.

sparerd:
I think for now I'll keep everything else as-is. The only part that I'm still considering is making save on edit default to enabled. It's probably the right move going forward, but will this catch some users by surprise? I don't want it to cause issues if a user can't have it enabled.

The one example I can think of is for massive connections files. In my last job I had a connection file with 15k+ connections in it. Saving that file to local disk took ~3-4 sec. Creating and editing a bunch of connections would cause a lot of activity that could have been batched up all at once in intervals.

Could maybe just put a disclaimer in the release notes that it is being defaulted to enabled?

kmscode:
hmmm... Yeah, release note update would be good.

Connection saves are async, so there shouldn't be any noticeable performance impacts... Or  do you think that we might lose data by queuing up changes?

If we think large connection files are more common place, this is would really increase the need to switch from xml to a DB (sqlite for local storage) - so maybe we leave this as is and going forward try to focus on #182

Issue #681, opened 2017-08-16
New icon more noticable on windows black panel


anuar45:

kmscode:
Anyone good at Photoshop, etc...?

Looks like this is the "source" file for the icon:
https://github.com/mRemoteNG/mRemoteNG/blob/develop/mRemoteV1/Resources/Other%20Graphics/mRemoteNG%20-%20Icon.psd

But when I open it in Paint.Net (with the PSD Plugin) it doesn't look like our current icon... Best we have at the moment it seems... 

Starting from scratch would be fine too... 😄

sparerd:
The psd files can be opened via GIMP too

kmscode:
Resolved via #1232

PR #804, opened 2017-11-14
facelift files

evilbaschdi:
some images are already changed. I also provided the VisualElementsManifest.xml.
Because of missing compoments I were not able to build the whole UI to see all of the changes.

https://twitter.com/mRemoteNG/status/930464358190198785


sparerd:
Might be missing Citrix receiver? That's a common missing component. Here's the devenv setup doc if you need it:
https://github.com/mRemoteNG/mRemoteNG/wiki/Development-Environment

evilbaschdi:
that's correct. I hadn't installed it yet because I do not use it on my private computer.
Maybe I'll try it out later.

sparerd:
Np, thanks for the PR

farosch:
wouldn't it be more consistent, to use the same graphics as the mRemote-logo on the homepage?
e.g. something like this? (beware: quick and dirty)
2017-11-15 17_57_41-home _ mremoteng
I also think the two colors are a real distinct design element that people instantly associate with mRemoteNG and thus shouldn't get lost

Also these color would be better to fix issue #681

farosch:
Same thing for the header:
logo

kmscode:
@farosch You make good points, but I think the icons as submitted are still lighter than the current ones and would likely still address that PR.... Would need to play with it a bit.

@evilbaschdi Fantastic! Thanks!

Would it be possible to adjust header.bmp and Logo.png to have the text bigger and make better use of the space? Also for Logo, I think we'd still like the "Multi-Remote Next Generation" tag line. @sparerd what do you think?

Can you include any "source" files (psd, etc) within the PR?

@sparerd If I recall correctly, we need .bmp's for the installer. It doesn't accept other image formats, correct? Otherwise I would say we should standardize on png except where an ico is required.

farosch:
#806 Since I was bored, here's another PR so you guys can have a look at what I have in mind.

kmscode:
closing this PR in favor of #1232 which includes additional changes and source image files (PSDs).

PR #806, opened 2017-11-15
New Logo/Icons; Revised "About"-screen

farosch:
-set minimum Size for main form window to 800x600
-new Icon
-"About"-screen:
--changed logo seen to match mremoteng.org
--rebuilt layout of as a TableLayoutPanel
--removed maximum size value of form

2017-11-15 19_28_03-mremoteng - confcons xml

Icon in startmenu, on desktop and in taskbar:

new layout as seen in VS:
2017-11-15 19_28_21-mremotev1 running - microsoft visual studio


kmscode:
Looks good... Thanks!

We would need the additional *.png files modified as well. Do you have any "source" files (PSD, etc)?

I like both designs between this an #804. Both are modern and clean. 

@sparerd if we can get both PR's to include all necessary info to the point where they are both "equal" but different visually, I think we'll need do a poll and let our users decide on which icon set they like more.

farosch:
@kmscode @sparerd This was just a quick thing I did in GIMP. If I get a go for this design I will replace all files and provide PSD-templates.

What about not only letting users chose the Icon but maybe make a little competition about it? Let users have some time to send in their suggestions and later make a poll on the homepage? We could spread the word on social media...

farosch:
I was also thinking about something like this:
2017-11-15 20_29_53-unbenannt_ - paint 3d

Boxx1e:
This is what my creation looks like. Whipped it up last year
Icon

aplocher:
Just my $0.02: the logo looks killer... loves me some flat styling.  

Something about the icon seems a little off, maybe the white background.  Might not be a bad idea to have two styles of the logo, one for dark backgrounds and one for light backgrounds, and then perhaps use an ALMOST black background (e.g. #1c1c1c) for the icon (similar to the current icon, only keep it flat and stylized how you have it!)  

In fact, to contradict my own previous statement, I don't think Windows design guidelines recommend using a background at all anymore.  Since tiles in the Start menu automagically (I think) provide a background color.  That might be a good reason to use your chunkier/bulkier 3D Icon (which looks killer, too, btw) with no background.

As you can see in this example, none of these icons are using backgrounds (with the exception, kinda, of Webstorm which has a stylized background [10th icon]): 

image

Check out this link.  They actually have guidelines for shadowing on 3D Icons along with shadowing for Flat Icons (search page for "Drop Shadow")

farosch:
closed in favor of #1232

PR #829, opened 2017-12-06
Fix for connecting to azure instances (cloud services). LoadBalanceIn…

sirLoaf:
This is a fix for connecting to azure instances (cloud services) that require a LoadBalanceInfo Cookie

### DIFF
--- a/mRemoteV1/Connection/Protocol/RDP/RdpProtocol.cs
+++ b/mRemoteV1/Connection/Protocol/RDP/RdpProtocol.cs
@@ -3,6 +3,7 @@ using System.Collections;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
+using System.Text;
 using System.Threading;
 using System.Windows.Forms;
 using AxMSTSCLib;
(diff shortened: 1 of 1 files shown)


sparerd:
What do you think about having an option to set how the LoadBalanceInfo is sent? Maybe a checkbox in Tools -> Options -> Advanced called "Format LoadBalanceInfo for Azure"? This way the normal functionality stays intact (in case it breaks other cases) but those who need this set for Azure connections can make it work.

Over time we might be able to phase out the option if we can show that there isn't any downside to formatting the property this way.

If that works for you, I can accept this PR as is and will add the necessary setting. It can be included in v1.76 since we aren't too far into testing and this should be a low impact change.

sirLoaf:
Sure, makes sense to have an option for this. Thank you

sparerd:
Merged - thanks for the PR @sirLoaf!

sparerd:
Added option for enabling utf8 encoding of the rdp "load balance info" property in commit 8aeea4d

Starting in v1.76 alpha 2, this can be enabled in Options -> Advanced -> `Use UTF8 encoding for RDP "Load Balance Info" property`

While it might have made more sense to make this a per-connection property rather than app-wide, it is significantly less work to do it this way. With any luck, encoding it as UTF8 doesn't break any other systems that accept the load balance info property. That, or Azure starts accepting UTF16 encoded strings...

PR #913, opened 2018-03-09
Update .github templates

benwa:
Bumping up the quality of the templates with help from https://www.talater.com/open-source-templates

I figure more info is better than less

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,16 +1,30 @@
-<!--
-Only file GitHub issues for bugs and feature requests.  All other topics will be closed.
+<!--- Provide a general summary of the issue in the Title above -->
 
-Before opening an issue, please search for a duplicate or closed issue.
-Please provide as much detail as possible for us to fix your issue.
--->
+## Expected Behavior
+<!--- If you're describing a bug, tell us what should happen -->
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
 
-<!-- Bug -->
-|Detail|Value|
-|--:|---|
-|Operating system | Windows 10 x64 |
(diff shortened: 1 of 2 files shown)


sparerd:
Thanks for adding this @benwa! 

My comments didn't save from a few days ago.. i could have sworn they posted... :(
I don't remember all my comments so they must not have been all that important. Here's the one I do remember though:

What do you think about adding a quick description under each field? I saw it on an issue for either mstest or vstest recently though now I can't find it. It was a nice bit of extra prompting

Like:
# Steps to Reproduce
> Provide steps to reproduce the issue. Please be as detailed as possible.
1.
1.

benwa:
The descriptions are there, but they're wrapped in comments. (check [line 16](https://github.com/mRemoteNG/mRemoteNG/pull/913/files#diff-1a1c3dd142f76a5fad803a0c52839881R16))  That way it will show for the submitter, but won't display on the submission.

Do you want them visible in the actual submission too?

sparerd:
Ahh, I was looking at it in the rich-text view and missed all of that completely. I like it

benwa:
Awesome.  If you have a few minutes, run through https://www.talater.com/open-source-templates. I answered to the best that I could and did slight tweaking to the output but want to make sure it's all correct for this project.

sparerd:
The template on that site has a description field at the top. What are your thoughts on adding it to our issues template?

```
## Description
<!--- Provide a more detailed introduction to the issue itself, and why you consider it to be a bug -->
```

benwa:
I think that one is if you chose bug only. I put bugs and feature requests. They're all over [here](https://github.com/TalAter/open-source-templates/tree/master/site/templates/issues)

PR #935, opened 2018-04-05
Update Language es

marcomsousa:
Fix a typo in Spanish language


sparerd:
Thanks for the PR @marcomsousa! Since this is very low risk, it will be available in the next v1.76 pre-release

PR #938, opened 2018-04-06
Small visual correction of the forms

sli-pro:
- for the main form set the minimum resolution 1140; 603.
- for the form of port scans, done visual editing of components.


sli-pro:
before


become

sparerd:
Looks good @sli-pro! I like the layout changes you are proposing. This seems like a fairly low-risk change, so it will be available in the next v1.76 pre-release.

sli-pro:
Thanks!

Issue #940, opened 2018-04-07
Error in the translation of certain menu items.


sli-pro:
Hello!
In version mRemoteNG-Portable-1.76.3.32991 there was an error translating certain menu items. Any language is selected by default, and the translation will be English only.

![1](https://user-images.githubusercontent.com/38104048/38459130-aabab3e6-3aae-11e8-990f-a8720c1e8e6b.png)
![2](https://user-images.githubusercontent.com/38104048/38459134-af57d5d2-3aae-11e8-9308-93f2bd1df5bf.png)
![3](https://user-images.githubusercontent.com/38104048/38459137-b27067c0-3aae-11e8-8435-cc0e07af407d.png)
![4](https://user-images.githubusercontent.com/38104048/38459140-bedaca1e-3aae-11e8-8d3b-53d4cc535236.png)
![5](https://user-images.githubusercontent.com/38104048/38459142-c1c7d2ee-3aae-11e8-9fc8-0a024b52bc66.png)

sli-pro:
The file/View/Tools/Help language does not change on the msMain component.

           // 
            // mMenFile
            // 
            DropDownItems.AddRange(new ToolStripItem[] {
                _mMenFileNewConnection,
                _mMenFileNewFolder,
                _mMenFileSep1,
                _mMenFileNew,
                _mMenFileLoad,
                _mMenFileSave,
                _mMenFileSaveAs,
                _mMenFileSep2,
                _mMenFileDelete,
                _mMenFileRename,
                _mMenFileDuplicate,
                _mMenFileSep4,
                _mMenReconnectAll,
                _mMenFileSep3,
                _mMenFileImport,
                _mMenFileExport,
                _mMenFileSep5,
                _mMenFileExit
            });
            Name = "mMenFile";
            Size = new System.Drawing.Size(37, 20);
            Text = Language.strMenuFile;

sli-pro:
can be closed, fixed in v1.76 Alpha 6

PR #942, opened 2018-04-10
Made several strings translatable and added Russian translations

sli-pro:


sparerd:
Github didn't add a reference to the issue since it was in the title. Adding it here so they are linked #940

sparerd:
Thanks for the PR @sli-pro! They will be included in the next v1.76 release

sli-pro:
Thanks!

PR #944, opened 2018-04-10
top-most on tray double-click

farosch:
fixed window not automatically being top-most when double-clicking tray icon


farosch:
I think the first connection after filtering should also be automatically selected - unfotunately I am unable to do this myself

sparerd:
Even though this is only a 2 line change, I'm going to schedule this for v1.77. This will need some testing to ensure it does not have a negative impact on other related actions.

sparerd:
I've created #1089 to talk about the connection filter enhancement more. I think we will need a more involved solution to get it working like users would expect.

PR #948, opened 2018-04-14
Translation

farosch:
some changes regarding tranlsation


sparerd:
Since this is just translation work, it will be included in v1.76

sparerd:
Changed target to fix some merge conflicts caused by another PR. Thanks for the translation work @farosch!

PR #1016, opened 2018-07-02
Chinese simplified language update for latest commit 05c96da

wwj402:
Only chinese simplified language update.


kmscode:
@sparerd My Chinese is a little rusty... But I'm OK with this PR... 😄

sparerd:
Thanks for the translations @wwj402!

PR #1107, opened 2018-09-22
Implement HiDPI support

### LINKED ISSUE #321: High DPI Support
Computers using HiDPI windows scalling with non hidpi apps will blur the interface (it will scale the interface like a bitmap, not like a vector).

To add HiDPI Support, just add this to manifest file.
VS2015 already have it, just need to uncomment the section.

`
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" 
          xmlns:asmv3="urn:schemas-microsoft-com:asm.v3" >
 <asmv3:application>
   <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">
     <dpiAware>true</dpiAware>
   </asmv3:windowsSettings>
 </asmv3:application>
</assembly>`

source: http://stackoverflow.com/questions/36336797/do-i-have-to-handle-hidpi-capabilities-within-my-source-code-or-is-it-provided

sparerd:
## Description
Implements HiDPI support which allows for proper window and control scaling when scaling is over 100%. This will allow hi-resolution screens to show mRemoteNG clearly (without blurring).

## Motivation and Context
Resolves #321

## How Has This Been Tested?
Visual checks on a hi-res screen and comparison with v1.76 to ensure layouts are comparable.

## Types of changes

## Checklist:


sparerd:
I implemented the hidpi changes in the UI by setting scaling to DPI, but I'm starting to wonder if it would be better to scale by Font. I don't think DPI scaling is affected by the scaling applied within Windows display settings (scale by 125%, 150%, etc). Vision-impaired users that just need to crank up the display size should also be considered by the implementation. I'll need to test this to see if a change is required.

sparerd:
Hmm after a quick test I can't tell the difference between Font scaling and Dpi scaling. I think I'll leave it like it is. If we get reports of display issues we can revisit this.

popofe:
@sparerd , It would be a shame to stay there while you are so close to the goal. I saw that after adding the second pair of parentheses in the line below, it worked fine on my screen with a scalefactor of 150.

`var scaleFactor = (uint)(_displayProperties.ResolutionScalingFactor.Width * 100);`

Issue #316, opened 2016-12-20
PCB (VMid) option for RDP Remote Desktop


ethancedrik:
In clients like FreeRDP you can specify a PCB (preconnection blob/VMid) if you are connecting to a HyperV console session. I think this feature should be implemented in mRemoteNG. 

Here is some more info on how that can be done and implemented in an existing/in development Remote Desktop client, https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/6527eebb-e2db-4607-8ab0-be3d16b5f1cd/rdp-preconnection-blob-settings-on-windows-server?forum=windowsgeneraldevelopmentissues

vincentparrett:
+1

sparerd:
We are doing some RDP enhancements in v1.77. This could probably be squeezed in at the same time.

S-dn-Y:
Hi Guys,

I really love mRemoteNG, this feels like the only thing that is missing.
I see there's an active pull request on this.
Is this still scheduled in the 1.77 release and/or when is the update (or this functionality) expected to arrive :)

Appreciate the effort!

Best regards.

sparerd:
It is very likely we will be able to include this with v1.77. We are still in the development stages for this release and don't know when this particular feature will be included

S-dn-Y:
Alright, thanks for the reply David.
I'm looking forward to it!
Have a good one, cheers!

PR #1176, opened 2018-11-02
Support for Hyper-V VM connections

popofe:
Allows connection to Hyper-V VMs using the VM Id

## Description
Add an entry for entering the VM Id in the connection tab
Add an entry to enter the Enhanced Session Boolean in the protocol tab
Add the "RDP on VMBus" protocol in the protocol tab

The use of the VM Id instead of the VM Name is due to a strange choice of Microsoft which imposes to have elevated privileges to make the association between these two notions. In addition, RDP only works with the VM Id. So, if we want to be able to use mRemoteNG without being an administrator, we have to go through the VM Id

## Motivation and Context
#316

## How Has This Been Tested?
Test on Windows 10.0.17134.345 

image

## Types of changes

## Checklist:


popofe:
Too many problems with jenkins. I will make a new Pull Request after cleaning these errors.

PR #1179, opened 2018-11-04
Support for Hyper-V VM connections

popofe:
Allows connections to Hyper-V VMs  using the VM Id

## Description
Add an entry for the VM Id in the connection tab
Add an entry for the Enhanced Session Boolean in the protocol tab (allow console mode or Enhanced mode for Microsoft OS) 
Add the "RDP on VMBus" protocol in the protocol tab

The use of the VM Id instead of the VM Name is due to a strange choice of Microsoft which imposes to have elevated privileges to make the association between these two notions. In addition, RDP only works with the VM Id. So, if we want to be able to use mRemoteNG without being an administrator, we have to go through the VM Id

## Motivation and Context
#316

## How Has This Been Tested?
Test on Windows 10.0.17134.345 

image

## Types of changes

## Checklist:


nmat:
Any news about this? I am curious since I have lots of hyper-v testing machines on my system :)

sparerd:
I've been meaning to test this for a long time now, but haven't got around to setting up hyper-v. Would you be willing to test it out @nmat?

nmat:
I can test it out. If you want @sparerd let me know if you need any specific feedback or so...looks like the pull request is kind of behind on the current development cycle.

I will give it a try tonight.

sparerd:
Things to look for:
- Does it connect to a hyper-v host?
- Fiddle with a bunch of connection properties, do they seem to work properly?
- Does it make sense to have a `RDP on VMBus` protocol, or could we get away with merging these new fields into the normal `RDP` protocol? What is the user experience of setting up a connection to one of these hosts?

sparerd:
I cleared up the merge conflicts - though let me know if anything looks off.

nmat:
Feedback:

I have not managed to connect to host.

This is good I think it can be improved a bit.

1. Get-VM | select VMName,VMId - this part should be done on request if you choose for example protocol to use RDP on VMBus (if it finds any VMs then it should list them auto with vmname and have the id listed as a extra information)
2. Hostname/IP - should not be required
3. Login is trying to use [mycomputer]\Username which should be \Username only

At the moment its quite a effort to get started with even testing to connect. So it might be good for beginners to have a bit more automation to the merge.

sparerd:
@nmat What is the full list of steps that a user would go through to get an entry added and working in mRemoteNG?

nmat:
To get started with this (even though I have not been able to connect properly using RDP on VMbus):

1. Change protocol
2. Add "localhost" to Hostname/IP
3. Get the vmid as administrator in powershell with: Get-VM | select VMName,VMId
4. Add it to the VM ID part of properties
5. Try to connect

Also tested 2 different accounts + manually typing it in when the question for auth comes up.

Normal RDP works as intended so something else is going on.

Now I tested with no username or password in order to see if I could fill one in and I got

image

nmat:
Thinking one extra time on the hostname......I wonder if this could be if there is a remote server running hyper-v to connect to that server and use that server vm bus ID....

popofe:
I agree that the solution I implemented is not very ergonomic. This comes from the constraints that Windows imposes. Here is the problem.

The cmdled "Get-VM" requires being an administrator. So there are two possible approaches.
- Either one authorizes the entry of the name of the VM but that supposes that mRemoteNG have elevated privileges to be able to find the VMId.
- Either we want to continue to be able to launch mRemoteNG by being simple user and we have no choice: we need to know the VMId.

I preferred the second solution. @nmat, @spared: what do you think?

As for the credentials, it seems to me that it is not possible to transmit them via VMBus. I would have to disable the fields Username, Password and Domain. Do not put anything in these fields.Get-VM | select VMName, VMId
@nmat, the error you have "An internal error has occured" comes from the fact that you do not have elevated privileges to connect to your VM (mRmoteNG has not been started as administrator or you are not in the hyper-v admin group). It is therefore necessary to authorize access to the VM in advance with the command:
`Grant-VMConnectAccess command -VMName <VMName> -UserName <UserName>`

Here is the list of step that a user would go through:
To do once with PowerShell with elevated privileges (or hyperv admin)
- Retrieve the VMId with the `Get-VM command | select VMName, VMId`
- Allow connection to the VM with the `Grant-VMConnectAccess command -VMName <VMName> -UserName <UserName>`

In mRmoteNG
- Enter the hostname
- Enter the VMId
- Select VMBus protocol

@spared, I'm not very comfortable with Git. I stupidly destroyed and recreated the develop branch of my repo and AppVeyor now generates the error "Fetching repository commit (142bcb7) ... Can not download GitHub repository contents: InternalServerError". Can you help me restore the situation?

sparerd:
> - Either one authorizes the entry of the name of the VM but that supposes that mRemoteNG have elevated privileges to be able to find the VMId.
> - Either we want to continue to be able to launch mRemoteNG by being simple user and we have no choice: we need to know the VMId.

I agree that approach 2 is the way to go, at least for now.

I've re-run the appveyor build and it looks like everything is ok. Does your git branch look ok?

sparerd:
I've been thinking about this more and I think it makes sense to have an RDP property rather than a new protocol for setting up connections over VM bus. We can still control the display of properties in the config window to help users configure their connections.

My proposed changes would be:
- Remove `RDPonVMBus` from `ProtocolType`
- Add a `bool` connection property for RDP connections. Something like "Use VMBus"
- It sounds like we should also hide the user/domain/password fields for these connections, but only if they are *never* needed for this connection type.

@popofe @nmat Thoughts?

mRemoteV1/Config/Serializers/ConnectionSerializers/Xml/XmlConnectionsDeserializer.cs
```diff
@@ -531,6 +531,10 @@ private ConnectionInfo GetConnectionInfoFromXml(XmlNode xmlnode)
                     connectionInfo.RDPAlertIdleTimeout = xmlnode.GetAttributeAsBool("RDPAlertIdleTimeout");
                     connectionInfo.Inheritance.RDPAlertIdleTimeout =
                         xmlnode.GetAttributeAsBool("InheritRDPAlertIdleTimeout");
+					
+                    connectionInfo.VMId = xmlnode.Attributes["VMId"].Value;
+					connectionInfo.EnhancedSession = bool.Parse(xmlnode.Attributes["EnhancedSession"].Value);
+                    connectionInfo.Inheritance.EnhancedSession = bool.Parse(xmlnode.Attributes["InheritEnhancedSession"].Value);
```
sparerd: These lines should use the `GetAttributeAs...()` extension methods to safely retrieve the attribute values.

```csharp
connectionInfo.VMId = xmlnode.GetAttributeAsString("VMId");
connectionInfo.EnhancedSession = xmlnode.GetAttributeAsBool("EnhancedSession");
connectionInfo.Inheritance.EnhancedSession = xmlnode.GetAttributeAsBool("InheritEnhancedSession");
```

mRemoteV1/Connection/Protocol/RDP/RdpProtocol.cs
```diff
@@ -157,8 +157,26 @@ public override bool Initialize()
...
                 {
-                    _rdpClient.AdvancedSettings7.EnableCredSspSupport = _connectionInfo.UseCredSsp;
+                    _rdpClient.AdvancedSettings7.EnableCredSspSupport = _connectionInfo.Protocol == ProtocolType.RDPonVMBus ? true : _connectionInfo.UseCredSsp;
                     _rdpClient.AdvancedSettings8.AudioQualityMode = (uint)_connectionInfo.SoundQuality;
+
+                    if (_connectionInfo.Protocol == ProtocolType.RDPonVMBus)
+                    {
+                        MSTSCLib.IMsRdpExtendedSettings _settings = (MSTSCLib.IMsRdpExtendedSettings)((AxMsRdpClient8NotSafeForScripting)Control).GetOcx();
+                        object value = true;
+                        _settings.set_Property("DisableCredentialsDelegation", ref value);
```
sparerd: Use the `SetExtendedProperty()` class method here instead. It will save you some setup and handle errors gracefully.

mRemoteV1/Config/Serializers/ConnectionSerializers/Xml/XmlConnectionsDeserializer.cs
farosch: Fixed by 39233d2

mRemoteV1/Connection/Protocol/RDP/RdpProtocol.cs
farosch: Fixed by 9128928

farosch:
@popofe 
This pr will be closed in favor of #1422. We decided to not implement a new protocol for this feature but instead make it an extension to the rdp protocol.
Therefor it is easier to just create a new pr instead of modifying this one.
We will definitively reuse some of the code you wrote and all the information you provided was very helpful for us, too. So please don't see this as an discouragement! We very much appreciated this pr and would love to see you contributing to this project again in the future!

Thank you very much for the code and information you provided! 👍 

You will very likely see this feature in v.1.77.
We are planning to release an alpha build in the upcoming weeks!

PR #1208, opened 2018-12-12
Added show parents name on tab name option.

onexey:
## Description
Added option under Tools>Options>Tabs & Panels named "Show parent information on tab names"

## Motivation and Context
When multiple parent-sub style connections exist with same name creates a confusion about which PC to connect.

## How Has This Been Tested?
Tested manually with different parent levels.

2018-12-12 18_49_29-mremoteng - confcons xml
2018-12-12 18_49_48-mremoteng - confcons xml - epdi-netpc01

## Types of changes

## Checklist:


auchri:
I think it would be better to not make this option and to only show the folder name when two connections with the same name are opened, as it is done by many other programs.

sparerd:
I agree with adding a sub-option for only showing hierarchy info on the tab if a duplicate item is open.

> Open tabs check may arise some questions such as case sensitivity, white space handling etc.

This seems reasonable:
- White space is counted. `host 1` != `host1`
- Not case sensitive `host1` == `Host1`

kmscode:
@sparerd - My concern with this PR is that we had length complaints in the past. Reference:
07ed1b3f741ae97bf28b4dd524cb373e99e82c85

https://github.com/mRemoteNG/mRemoteNG/blob/develop/mRemoteV1/Connection/Protocol/Http/Connection.Protocol.HTTPBase.cs#L186

https://mremoteng.atlassian.net/browse/MR-586

But - as long as this is not enabled by default, I'm OK with it...

onexey:
That's good to know.
I won't be around till end of the January but after that I'll change to code as follows:

- Add option to show parent name on title.
- Add sub option to only show parent name if there is duplicated connection.

I'll be happy to hear any suggestion/opposition till then.

sparerd:
Hey @onexey, just checking in. A recap of the requested changes:
- Add a sub-option to only show the hierarchy if a connection with a duplicate name is open. (Option disabled by default)
- Change the default of `ShowParentInfoOnTabs` to false to retain the current behavior of the app

onexey:
Yes that is correct. I'll make the changes in near future.

Kvarkas:
I am having some ideas how to make this a bit more workable and visible.

Kvarkas:
after careful consideration I decide not to implement that due coming changes in profiling and new UI this seems to be absolete

Issue #1216, opened 2018-12-20
Request: Move searchbar to top of connection tree


farosch:
The searchbar in the connection tree panel is currently placed at the buttom of the window.
The handling would IMHO be easier, if placed at the top for numerous reasons. Some of them being:

- Mostly searchbars are placed at the top, so this is where you'd expect it to be
- All functional UI elements are gathered at the same place
- Shorter distances to travel with your cursor, especially when filtering (see screenshot below)

![image](https://user-images.githubusercontent.com/22905106/50284560-aa767080-0459-11e9-86b9-7ae0f91e63cf.png)

This is a simple mock-up of how it could look:
![gimp-2 10_2018-12-20_13-16-21](https://user-images.githubusercontent.com/22905106/50284595-cc6ff300-0459-11e9-9c64-2ae4d8a6f95e.png)

I'd be happy to provide a PR if you want me to ;)

kmscode:
I would love a PR for this! I think it makes a lot of sense.

PR #1217, opened 2018-12-20
Moved searchbar in connection tree to top

farosch:
See #1216 
This is what it looks like:
image

### DIFF
--- a/mRemoteV1.sln
+++ b/mRemoteV1.sln
@@ -73,6 +73,7 @@ Global
 		{1453B37F-8621-499E-B0B2-6091F76DC0BB}.Release|x86.ActiveCfg = Release|x86
 		{1453B37F-8621-499E-B0B2-6091F76DC0BB}.Release|x86.Build.0 = Release|x86
 		{5423D985-CB48-4344-B47F-E8C6D60C8B04}.Debug Portable|Any CPU.ActiveCfg = Debug|x86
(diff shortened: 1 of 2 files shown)


sparerd:
I think this is sensible - merging

nmat:
Hmmmm shouldn't this be configurerable instead? For my eyes....it unfortunately gets cluttered in the top bar. :/

farosch:
I wouldn't call five elements in two rows cluttered, but I think this should be fairly easy to implement if desired.
I initially just wanted to clean the UI a little bit without introducing more lines of code, but I guess I could give it a shot. @sparerd What do you think?

nmat:
Oh I agree with the change. For _me_ its easier with it at the bottom. So I am only thinking about if it could be configurable. :)

sparerd:
Usually I don't mind adding configuration options, though this one I'm on the fence about.

@nmat If the search bar was more visually distinct from the buttons, would that improve things? Or does it just feel weird that the search bar is at the top?

sparerd:
That's fair. We can make it an option with the default set to bottom

farosch:
That's how it looks atm

sparerd:
That looks good - would you mind opening a new PR to add the config option?

nmat:
thats perfect! thank you!

PR #1227, opened 2018-12-22
Position of search bar configurable via settings

farosch:
As talked about in #1217


fxrsliberty:
as a daily user, actually my entire day. I would really like the "connection " bar to search my connections before initiating a new connection with the default protocol. 
i.e. I have an RDP connection configured as ad1.mycompany.com but the default protocol is ssh... since I only use the AD server to add users to the network..... if the connection bar would show the connection name as I type for the configured connection list . OMG, my life managing 300+ lunux boxes would be so much faster....

PR #1223, opened 2018-12-21
Open external links in browser and remove internal help

farosch:
Remove internal help coded into mremote and open links to website, donate, forum and bugs in the users default browser instead of mremote itself. See #947

## Description
image
These links on the help file menu now open in a browser instead of mremote itself.
All help files and the needed class to display them were deletet from the source code

## Motivation and Context
 - Help files inside of mremote were outdated and had screenshots of very old versions of mremote in them
- Websites were viewed in a webbrowsercontrol which uses IE. This is unsafe and some pages (eg github) don't support IE anymore
 - People that are willing to contribute to the documentation cannot do so without at least some experience with Visual Studio
 - Redundancy. Help files need to be cared of at two seperate locations (or should have been)
 - Flexability. Update/extend/add help files without the need to provide a new release
 - Maintainability. One thing less to care about
 - Usability. Bookmarks, auto-login, etc.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.

PR #1228, opened 2018-12-22
External links

farosch:
- External links will open in the default webbrowser instead of inside mremote
- Changed links to https

Replaces #1223


kmscode:
Thanks for the PR. Merged!

Issue #785, opened 2017-11-08
Texts from Settings seem to be hard-coded and are thus not translatable


farosch:
Hopefully I am not overlooking something here but to me it seems like the following fields can not be translated through a resource file.

![1](https://user-images.githubusercontent.com/22905106/32577860-8db70144-c4db-11e7-9817-c0361d73fa19.png)
![2](https://user-images.githubusercontent.com/22905106/32577861-8dd389e0-c4db-11e7-957d-278f3feb9270.png)

kmscode:
Off the top of my head, I believe you are correct. These are newer ui components so we might have inadvertently skipped the localization bits. Will look to get them into the EN resource asap.

PR #1229, opened 2018-12-23
Language fix

farosch:
See #785. (The string in the second picture was already is translated by now)
The text containing the update channels explaination was a standard TextBox, changed it to a NGLabel instead.


farosch:
I moved the reconnect button from the tab context menu to the first position. It felt a little off in the middle of the tab handling options:
**Before:**
devenv_2018-12-24_21-31-21

**After:**
devenv_2018-12-24_21-31-59


Also "Disconnect Other Tabs" only shows up when there are actually other tabs and
"Disconnect Tabs to the right" only shows up when there are tabs to the right

sparerd:
> I moved the reconnect button from the tab context menu to the first position. It felt a little off in the middle of the tab handling options:

It also feels weird having Reconnect/Disconnect in different places. "Duplicate Tab" is also essentially just "Duplicate Connection", which fits the theme of connection state management. The only one that doesn't really fit with that logic is Rename Tab.

sparerd:
I'll concede that Duplicate and Rename Tab probably aren't used that often. Another option would be to move both the reconnect/disconnect stuff to the top.
```
Reconnect
Disconnect
Disconnect Other Tabs
Disconnect Tabs To The Right
----
Fullscreen (RDP)
...
```

sparerd:
Though changing the location of those items really doesn't fit within this PR, since this is for fixing some localization issues.

farosch:
Should I undo changing the icon position and make a new PR for this?

sparerd:
Yes, that might be best

farosch:
Closed, see #1229

PR #1231, opened 2018-12-23
Tracks the active tab in the treeview

### LINKED ISSUE #1230: Tracking of the selected tab in the treeview
This is to track the implementation of tracking the selected tab in the treeview. I'll submit a PR against this issue soon.

SeanTMill:
Fixes #1230. I didn't implement an option for this feature because I didn't want to delay it for translation and I wasn't sure of the best way to insert a new option. I don't know if it is desired to insert it into the existing list somewhere around another similar option or just append it to the bottom.

### DIFF
--- a/mRemoteV1/UI/Forms/frmMain.cs
+++ b/mRemoteV1/UI/Forms/frmMain.cs
@@ -559,6 +559,7 @@ namespace mRemoteNG.UI.Forms
 			{
 				titleBuilder.Append(separator);
 				titleBuilder.Append(SelectedConnection.Name);
+                Windows.TreeForm.JumpToNode(SelectedConnection);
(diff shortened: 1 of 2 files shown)


sparerd:
I've tested the changes and they seem to work well. We will want to add an option for this functionality.

> I don't know if it is desired to insert it into the existing list somewhere around another similar option or just append it to the bottom.

We typically group up options that do similar things, so I would insert this option between the `Single click on opened connection in...` and `Set hostname like display name when...` options in the mRemoteNG.UI.Forms.OptionsPages.ConnectionsPage class.

Let me know if you need help with any of this

sparerd:
Merging - the user option will be easy enough to add

Issue #765, opened 2017-11-02
Port Scan Issues


arrrghhh:
Using Port Scan feature, I want to only scan port 3389.  I put '3389' as start/end port, yet system still seems to scan all of the default ports.

Also, the port scan itself when I changed the range it did not update properly.  It would still scan the old range, I had to close the tab and open a new port scan tab in order to scan a new IP range.

I find using PortQry.exe is quite fast (assuming the port is open) so I would imagine this would speed up the port scan itself quite a bit if it only focused on one port.
|--:|---|
|Operating system | Windows 10 x64 |
|mRemoteNG version| 1.75.7008  |

sparerd:
Relevant discussion:
https://www.reddit.com/r/mRemoteNG/comments/7a4nwi/port_scan/

PR #1241, opened 2018-12-27
Two issues fixed

### LINKED ISSUE #1044: Dragging (grabbing) the program window requires 2 clicks.
## Expected Behavior
With every other windows program, windows can be grabbed and dragged by one click then drag.  

## Current Behavior
However, with mRemoteNG the first click only seems to highlight the window but not give the ability to drag it.  To drag you have to click the window a second time!  It sounds like a small thing but it's really annoying.

## Steps to Reproduce (for bugs)
Click one on the mRemoteNG window coming from another context (ie from using a different windows program, and not while already using mRemoteNG)
Try to drag the window (you won't be able to).
If you click again, you can now drag the window.

## Context
If mRemoteNG worked as every other program, workflow would be improved as you wouldn't constantly make the failed attempt at dragging/moving the window.

## Your Environment
1.76.4.16229
Windows 10 Home, version 1803, build 17134.165

farosch:
This fixes #1044 and #765
I still have one problem. I just found out that calling thread.abort on the port scanner does not really abort the thread. Any ideas on why that could be? @kmscode


kmscode:
I think I figured out the port scan thread issue. I'll have to test a bit
more tonight or tomorrow morning. I'll push a fix to the PR if I'm right.

Thanks!!!

kmscode:
@farosch Thanks for the fixes. Merged. I just did some minor tweaks after testing. Also best attempt to cancel the scan - the pings get queued up so fast that canceling them isn't guaranteed. but with this change at least it looks like it gets cancelled and results won't be posted to the UI. Pings may still be running in the background until they complete however.

PR #1243, opened 2018-12-28
rearranged items and added/replaced icons

farosch:
See #1229 
devenv_2018-12-28_18-06-16
I changed the icons on the tab-related menu items since the pause button was kinda unfitting and also added icons for the putty settings and the vnc view-only option


kmscode:
merged develop (which has magiclib removed) into this PR and fixed some minor conflicts. OK to merge now.

PR #1252, opened 2019-01-05
Added support for RDP and ICA connections for password retrieval via AdmPwd.E.

jformacek:
## Description
AdmPwd.E ([www.admpwd.com](http://www.admpwd.com/downloads)) specializes on secure automatic password management and retrieval for local and domain accounts in Windows environment. Managed password is securely stored in Active Directory, and eligible persons can retrieve it from there.

AdmPwd.E also provides integration SDK library that allows password retrieval logic integration to other applications. This allows to eligible people getting and using password on the fly, without storing it, typing it, or even knowing it.  
This feature substantially increases security of local and domain accounts, because:
* password is automatically regularly randomized
* password does not need to be stored anywhere else than in Active Directory
* AdmPwd.E toolset allows for automatic password retrieval and usage

## Motivation and Context
Possibility to automatically retrieve connection password in RDP manager greatly increases its value because:
* users do not need to care about password expirations and changes to update passwords stored in mRemoteNG credential storage
* passwords do not leave the company - they aren't stored in connection file
* automatic password retrieval is fully integrated with current mRemoteNG features of default credentials to be used

User has possiblity to turn on automatic password retrieval when configuring default credentials in Options - Credentials dialog box. Nothing else is needed - just keep password field empty in Connection configuration, and mRemoteNG retrieves password when connection is established.

## How Has This Been Tested?
Feature was tested with and without option to use AdmPwd.E for password retrieval turned on. Without this option, mRemoteNG behaves the same as if the support for AdmPwd.E was not implemented - so user experience of users who do not want to use this feature is not impacted.

When option is turned on, and password retrieval succeeds, it was tested that connection is successfully established.

When password retrieval fails for some reason, it was tested that mRemoteNG just asks the user for password, the same way as if the password is not specified.

UI language translations for new UI element (checkbox) were added for neutral and Czech languages. It was tested that Czech translation displays when language of mRemoteNG is set to Czech; for other languages for whose translation was not provided it was tested that neutral language message is displayed.

image

## Types of changes

## Checklist:


farosch:
Replaced by #1662

PR #1276, opened 2019-01-13
Set focus to textbox instead of connecting, 

### LINKED ISSUE #1271: Feature Request: Change behaviour of quick connect toolbar
## Expected Behavior
50959116-80492900-14c2-11e9-8331-0eed1d87704e
 - Clicking the left part initiates a new connection
 - Clicking the arrow allows for changing the protocol without initiating a connection afterwards

## Current Behavior
 - As soon as the button is clicked or a new protocol is set mremote tries to initiate a connection

## Context
 - An error message is produces when changing the protocol with an empty hostname

 - After a connection is successfully initialized the hostname stays in the textbox. If I now want to open a new quick connection I'm forced to enter the hostname before chnaging the protocol type because otherwise it will instantly try to connect again using the new protocol

## Possible Solution
If possible create two tab stops (on the button itself and the arrow-part) to implement this without disturbing users that use keyboard shortcuts

## Your Environment

* Version used: Current develop branch
* Operating System and version: Windows 10 1809 x64

farosch:
This fixes #1271.
Changing the protocol in the quick connect toolstrip sets the focus to the textbox instead of directly connecting.
Users that use the keyboard can still initiate a connection when hitting `ENTER` afterwards.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/mRemoteV1/UI/Controls/QuickConnectToolStrip.cs
+++ b/mRemoteV1/UI/Controls/QuickConnectToolStrip.cs
@@ -203,7 +203,7 @@ namespace mRemoteNG.UI.Controls
         private void btnQuickConnect_DropDownItemClicked(object sender, ToolStripItemClickedEventArgs e)
         {
             SetQuickConnectProtocol(e.ClickedItem.Text);
-            btnQuickConnect_ButtonClick(this, e);
+            _cmbQuickConnect.Focus();
         }
 
         private void SetQuickConnectProtocol(string protocol)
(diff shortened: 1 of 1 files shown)


mRemoteV1/UI/Controls/QuickConnectToolStrip.cs
```diff
@@ -203,7 +203,7 @@ private void cmbQuickConnect_ProtocolChanged(object sender, QuickConnectComboBox
         private void btnQuickConnect_DropDownItemClicked(object sender, ToolStripItemClickedEventArgs e)
         {
             SetQuickConnectProtocol(e.ClickedItem.Text);
-            btnQuickConnect_ButtonClick(this, e);
+            _cmbQuickConnect.Focus();
```
kmscode: I suggest focusing the text box only if it's not filled out:

string.IsNullOrEmpty(e.ClickedItem.Text) ? _cmbQuickConnect.Focus() : btnQuickConnect_ButtonClick(this, e);

This would allow the proper workflow in either scenario...

PR #1332, opened 2019-03-01
Add show/hide main menu function to connection tree toolbar.

wangxuepeng:
## Description
Added shortcut function "Show/Hide Main Menu" under connection tree toolbar.

## Motivation and Context
Want to enlarge the content window to maximum while not want to enter full screen mode.

## How Has This Been Tested?
Tested manually with multiple windows and different languages.

screen shot 2019-03-08 at 10 35 21
screen shot 2019-03-08 at 10 35 29

## Types of changes

## Checklist:


kmscode:
@sparerd This look OK to merge after a quick review.  What do you think?

The Jenkins failure looks like to be a Jenkins problem, so nothing concerning there...

farosch:
@kmscode code-wise it looks ok but ui-wise it'll cause at least one problem:
image

farosch:
I would put an option like this into the windows application menu or whatever it's called:
mRemoteNG_2019-03-16_11-47-45


Also this should be changed from an option to hide the main menu into an option to hide the whole menu strip container

wangxuepeng:
> I would put an option like this into the windows application menu or whatever it's called:
> mRemoteNG_2019-03-16_11-47-45
> 
> Also this should be changed from an option to hide the main menu into an option to hide the whole menu strip container

This idea is awesome!!

nmat:
@farosch what is the status of pushing this to the next 1.77 build?

farosch:
@nmat I'd like @sparerd and @kmscode opinion on this before merging anything because I don't really like either of these two approaches to hide the main menu..

2pascal:
Even if the implementation isn't ideal, could a fork be compiled with these changes for the group of users that wouldn't mind?

farosch:
> Even if the implementation isn't ideal, could a fork be compiled with these changes for the group of users that wouldn't mind?
Yes, this is absolutely possible. mrng is an open source program and everyone can fork it and release their own modified version of it if they want to.
But we will never do this - A feature either gets implemented or it doesn't. There is one version of mrng only and it'll stay that way.

farosch:
Sorry for the long wait guys - merging atm...
Thanks again to @wangxuepeng for the original PR!

PR #1404, opened 2019-04-12
Correct the Chinese translation in https://github.com/mRemoteNG/mRemoteNG/issues/1403

carrtohu:
Correct the Chinese translation in https://github.com/mRemoteNG/mRemoteNG/issues/1403


sparerd:
Thanks for the fix!

PR #1430, opened 2019-04-30
Minor changes and forms of the Russian language

sli-pro:
mRemoteNG -1.77 offset components in the forms of inaccuracy in the Russian language.

fix


farosch:
Thank you very much @sli-pro!

PR #1433, opened 2019-05-01
Update icons in the Help section

sli-pro:
Update icons in the Help section.
If you like the icons, you can update them in all sections.
https://github.com/elementary/icons
------------------------------------------------------------
Fixed not correct the link https://mremoteng.org/contribute


farosch:
@sli-pro Thank you very much for this suggestion. The icon set of elementary OS looks really nice.
Unfortunately replacing every icon in the app is very time consuming and currently not very high on our priorities list.
We would have to check beforehand if the elementary OS icon set has all the icons mremoteng needs and we would very likely heve to create new ones by ourselves.
We would also have to make some changes to the code as using 16x16 icons is no good pratice anymore. Especially on high dpi setups these small icons often look blurry.
So we would need to render icons according to the display resolution from svg files or something inside the code.

If you would like to do so, I suggest you join us on gitter so that we can talk over how to implement a new icon set accordingly.

Issue #870, opened 2018-01-25
Enhancement - Push Inheritance from folder to all objects below


flyer292:
Having the ability to enable inheritance on all object within a folder would save a lot of time for updating existing connections.

Same type of feature that is available on windows file security when you replace all child object permissions with inheritable permissions from this object

sparerd:
Good idea. We will need to think about how this would work in the ui, but actual implementation shouldn't be too bad

JRVcr:
I'd like to see inheritance enabled by default for new objects.

malfuncion:
I was going to open this as a bug as it was working before updating to the latest beta. I have a set of connections nested below 1+ folders. So I have Top Level --> Client --> Service Account 1 and Service Acct 2 folders --> System Type  --> Groups of connections. The connection inherit from the system type folder, which then inherits from the SVC Account 1 and 2 folders above. In previous versions, this worked. With last update, no longer pushes down the tree. Noticed that I can set inheritance on the connection level and it works for that sessions only. Once I close the GUI and reload it is reverted on startup to "false" in the confCons.xml. Shutdown keeps the flags to "true". I also see an error about cannot load "donotplay" for object referring to sound passthrough. Possibly related, but not sure?

penguinairlines:
I am also looking for this feature.  Changes in an environment, feature ads, and importing a co-worker's session file all warrant this as a value-able feature.  

For example, once I migrated between 2 versions of mRemoteNG and found that a new feature had been added "minutes to idle", which did not inherit on any of my existing credentials. In order to use this feature, I have to select each session/folder and change the inheritance for this feature to Yes.  Having the ability to recursively edit an inheritance property from a certain folder downwards would be extremely useful here.  I can also see an argument for applying the same with the Inherit All value, in a situation where a bunch of servers with individual credentials become attached to a domain.

sparerd:
I have this one mostly implemented. It needs a bit more testing but should make it into v1.77

JRVcr:
Good work, David; thanks! Happy to see development pace is picking up lately, too!

--

Jeff Vandervoort
Crescent IT Systems

quokka-perth:
Hi Guys, I'm very much looking forward to propgating inheritance.... how close are we now?

Kvarkas:
> Hi Guys, I'm very much looking forward to propgating inheritance.... how close are we now?

I also would like to implement/fix that so may say in upcoming version should be ok

[...]

PR #1451, opened 2019-05-15
Added option to push inheritance settings to children recursively

sparerd:
## Description
This PR adds a few new features for quickly mass-applying settings to certain fields and also changes to some UI behavior around inheritance.

- Two context menu items have been added (in a sub-menu) as shown below. 
   - `Apply default inheritance` makes the currently selected node's inheritance match the default inheritance you have set up.
   - `Apply inheritance to children` takes the currently selected node's inheritance and updates all children to match it. This is recursive, so sub-folders and their children will also be modified.
- The inheritance button in the connection tree window is now enabled for connections/folders directly under the root connections node. Previously, if a node's parent was the root connection node, the inheritance button would be greyed out (disabled). Please note that children of the root inheritance node still cannot inherit values from the root node, even if inheritance for a property is turned on.

## Motivation and Context
Implements #870

## How Has This Been Tested?
- Added unit tests for pushing inheritance to children
- Manual testing for UI/UX changes

image

## Types of changes

- [ ] Updated translation

## Checklist:


sparerd:
@kmscode @farosch @nmat @jcefoli I'd like some feedback on the user experience of setting inheritance values using this new feature. There were a couple decisions I made which I'm not too sure about.

- Does it make sense to show the inheritance of children directly under the root node?
   - Now direct children of the root node show their inheritance. However, their settings don't do anything because we still don't allow inheritance from the root connections node. I'm worried this will cause confusion with users who set inheritance on a folder directly under root but then are confused why it doesn't do anything.
- Should we enable the inheritance button on the root connections node?
   - Like above, I'm worried this will cause confusion. Any inheritance settings applied on the root node are useless, except in the case where you want to push those settings to children.
   - When I'm using the new feature it feels weird that I *can't* push settings from the root node. If I want everything in my connections file to inherit the same values, I either need to manually set those value on children of the root node, or have a dummy "root folder" which I put everything inside of to do the same (that adds 1 level of indentation for no reason).
   - If we enabled setting inheritance on the root node, we would need to start saving that information somewhere. We currently don't serialize the root node (except for a few special properties).
   - Having inheritance on the root node feels like the concept of default inheritance, which would get confusing. The difference between "root" inheritance and "default" inheritance would be pretty small. We could make the root node's inheritance be the default inheritance (unify the two ideas).

padillah:
I know you didn't ask me but...
As I was writing my initial response I realized that we are inheriting two different things. One is the settings - Do we share the keyboard, Do we use the same port... things like that.

The other is the inheritance itself. Do we inherit the "Do we share the keyboard" property.

It appears to me those two sets of information flow in two different directions. The inheritance settings flow from top to bottom, parent to child. The values that get inherited flow from bottom to top - the child decides to pull the value from the parent.

Then this just becomes a type of "refresh" of the top-down settings.

... unless I'm wrong.

sparerd:
@padillah I hadn't thought of it that way before, but yes that is correct. Children must decide whether they will take their values from the parent, like turning on a valve and letting water flow downward (the water being connection settings in this analogy).

This feature will allow us to force inheritance settings on a whole branch of the connection tree at once. Rather than the question of "should-we-inherit" being just a bottom-to-top data flow, we can now set it top-to-bottom for certain portions of the tree.

farosch:
I think we should enable setting inheritance on the root node.
We could just serialize it like any other node. This would make inheritance more intuitive to handle imho.
Also, I don't think we should differentiate between root inheritance and default inheritance. These two should be concised. I can't think of any disadvantage and don't like the idea of dummy folders.

sparerd:
That will end up being a rather large change, since default inheritance is managed via the app settings at the moment and root inheritance would come from the connections file. I do agree with you, but it's going to add quite a bit more work.

stryker09:
So I downloaded the build done by CI and doing the right click Edit > Apply inheritance to children didn't replicate the credentials across to the child connections. Was this intended? It wasn't clear in the PR description.

Also agree to rename Edit to Inheritance as it properly reflects its function in the UI

sparerd:
@stryker09 Real property values like username/password won't be pushed to child nodes. Only the inheritance settings are pushed.

Given this:
```
folder1 (Inherit Username = true)
   |- con1 (Inherit Username = false)
   |- con2 (Inherit Username = false)
   L folder2 (Inherit Username = false)
      L con3 (Inherit Username = false)
folder3 (Inherit Username = false)
```

Running "Apply inheritance to children" on folder1 produces:
```
folder1 (Inherit Username = true)
   |- con1 (Inherit Username = true)
   |- con2 (Inherit Username = true)
   L folder2 (Inherit Username = true)
      L con3 (Inherit Username = true)
folder3 (Inherit Username = false)
```

In this example, `con3` will use whatever username is set in `folder1`. More specifically, `con3` asks its parent (`folder2`) for its username value and `folder2` asks its parent (`folder1`) for the value.

sparerd:
I'll rename the `Edit` sub-menu to `Inheritance` as suggested

sparerd:
I've made that change. Should we also put some tooltip text over those sub-menu options to explain what they will do?

farosch:
@sparerd Updating the help files should be enough. Tooltips are great, but as far as I'm concerned none of the other menu items have tooltips atm. We should keep information at one place, at least for now, so users don't need to search at different places, I think.

TimHandy:
Hi, is this a dead-end or is it still being actively worked on? I would LOVE this feature. It's my only annoyance with the software.

majkinetor:
Yeah, what is going on ? :) This is awesome feature.

JustSomeGuyNZ:
Any updates or progress?

srekp:
1.76.20 (2019-04-12):

> I think we should enable setting inheritance on the root node. We could just serialize it like any other node. This would make inheritance more intuitive to handle imho. Also, I don't think we should differentiate between root inheritance and default inheritance. These two should be concised. I can't think of any disadvantage and don't like the idea of dummy folders.
> 
> **EDIT:** I would also change the name of the menu entry to _Inheritance_. Calling it _Edit_ makes me feel like _Duplicate_, _Rename_ and _Delete_ should be under there, too. chrome_UmFgSMuo4y

I can not find this feature in version v1.77.1 neither in v1.76.20 
Is it only available in Nightly?

Issue #1488, opened 2019-06-12
MSI of Version 1.76 requires KB2574819 (not designed for my Win7 client)


DelPantalone:
## Current Behavior

KB2574819 is required. When trying to install KB2574819, the update tells me it is not suiteable for my client.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context

Unable to install the MSI for v176

## Your Environment

* Version used: 
v1.76
* Operating System and version (e.g. Windows 10 1709 x64):
Win7 Enterprise Service pack 1 Build17110

sparerd:
As a workaround, set the `IGNOREPREREQUISITES` flag of the installer as described here: https://mremoteng.readthedocs.io/en/latest/installation/install.html#extend-installer-properties

This will be fixed as a part of #826 (currently being worked on).

farosch:
Closed, as duplicate of #1029

Issue #826, opened 2017-11-30
Allow selecting RDP version to use when connecting


sparerd:
When creating an RDP connection record, I should be able to select the version of RDP protocol to use.

This provides several benefits:
- Application requirements can be relaxed a bit. The currently required KBs are only required to make RDP work correctly. If we have older RDP protocol versions available, those KBs may not be required.
- We are not using the newest RDP protocols. Allowing users to select the protocol version lets users on newer systems take advantage of bug fixes and new features while not forcing users on older or more restrictive systems to use old versions of mRemoteNG itself (just for older versions of the RDP protocol).

This comes with some challenge:
- How do we determine which RDP protocols are supported? COM searches? Something similar to the component check page?
- Some versions of RDP introduce new functionality. Someone using an older RDP protocol will need to know which settings are valid for their version and which are not. How do we handle this in a user friendly way? This must also work in a shared environment (db support, but also xml synced between multiple machines that may not have the same capabilities).

sparerd:
Basic work for this is done:
- Users can now select the `RDP Protocol Version` to use when connecting via RDP. The default is `Highest`.
- Updated the component check page to show all supported RDP protocol version on the machine. We only show an error if *no* versions are supported (very rare these days).
- The installer no longer requires certain KB updates be installed on Win7 machines. There have been several reports of issues installing on Win7 due to different combinations of KBs that provide the same necessary components but which aren't honored by the installer.

I am still trying to find a way to inform a user when a particular RDP property will not be used because the required protocol version isn't available. I would like it to display in the config window, but the WinForms property grid isn't very easy to modify.

farosch:
@sparerd How about just showing a warning in the notifications window?

sparerd:
@farosch Yea that could work. I'll give it a shot and see how it looks

sjuffe:
When will this become available to download?  I've downloaded beta for 1.77.0 but it is not an option to change the RDP Protocal version to highest.

sjuffe:
Any response to this?  I can't seem to get any response about this.

> When will this become available to download? I've downloaded beta for 1.77.0 but it is not an option to change the RDP Protocal version to highest.

Kvarkas:
will check

Issue #323, opened 2016-12-27
Wallpaper does always show, colour depth does not apply


nighty2k7:
Connecting from a Windows 10 machine to Win2016/10 always shows the wallpaper, also colour depth settings do not apply.

Using latest beta/alpha 1.75.6179.28241

|||
|--:|---|
|Operating system | Windows 7 x64 |
|mRemoteNG version| 1.75 aplha 3  |

meffect:
can confirm wallpaper issue, never tried color tho

kmscode:
Hmm.... I thought I saw some other reports of similar problems, but with different RDP options...

I'll look into this further. No ETA at this time.

nighty2k7:
Hi, any news on this issue?

To be honest, RDP with wallpapers, effects and not being able to use 16bit sucks performace-wise. Performance could be so much better if those parameters would be parsed correctly.

Cheers and thanks
Nighty

boomshankerx:
This appears to continue to be an issue. Any feedback on this?

ner00:
Can confirm this issue is still present.

boomshankerx:
It appears as though @kmscode may have gotten busy doing something else. Commits are down in 2019. Such is the life of the free software programmer. Maybe we should offer to throw some money at it by way of donations. I'm using MS RDP client on slow connections to Windows 10 for now. @sparerd seems to be active. Maybe he can shed some light on this. Some feedback either way would be appreciated.

ner00:
Yeah, I was only half-joking.

I don't think that throwing money at it is the way to go, but if that got this annoying and important bug to get fixed fairly quickly I would have to agree. As for the claim that I have read somewhere that this is exclusively related to Win10 hosts/guests, that's not true - the problem is the guest system (and by extension mRemoteNG client), I can say that during my testing this issue was confirmed on Windows 8.1, Windows Server 2012 R2, Windows 10 - probably everything in between up to Windows Server 2019; I haven't tested OSes older than Windows Server 2008 R2 (this one was the only where it worked).

I downloaded the source and played around with the variables, even though I don't understand the integration with RDP. In `RdpProtocol.cs` you have the `SetPerformanceFlags()` function which seems to be where the magic happens. 
```
                {
                    //pFlags += Convert.ToInt32(RDPPerformanceFlags.DisableWallpaper);
                    pFlags += 1;
                }
```
[...]

PR #1498, opened 2019-06-23
Add option to choose RDP protocol version

### LINKED ISSUE #489: Update 1.75.7005 to 1.75.7006 not possible - KB2574819 seem not be installed but it is
|||
|--:|---|
|Operating system | Windows 7 x64 |
|mRemoteNG version| 1.72.5065.32737  |

Update to actual Version not possible - Update package required but already installed. Update to actual Version and Version 1.74 not possible.

sparerd:
## Description
Implements #826 

A new connection property has been added (`RdpProtocolVersion`) which allows users to select the version of RDP protocol to use when connecting. At this time, the following values are supported: `Rdc6`, `Rdc7`, `Rdc8`, `Rdc9`, `Rdc10`, `Highest`. The default value is `Highest`, which dynamically selects the highest RDP protocol version supported on the client machine.

The component check page has been slightly redesigned for the RDP section. It will now check all RDP protocol versions used by mRemoteNG and report which ones are supported on the client machine. The machine will fail the component check for RDP if it does not support any protocol versions used by mRemoteNG (very unlikely).

In addition, this contains a fix for #323. We set the `NetworkConnectionType` to 1 to ensure the performance flags are not ignored.

## Motivation and Context
Prior to this change, we only supported RDC8. Making this a user setting enables us to support more recent RDP features without dropping support for older versions. In addition, we now no longer require certain Windows updates to be present on the machine during install (will resolve #489, #1488). 

## How Has This Been Tested?
Manually tested existing functionality during refactoring process to check for regressions.

image
image

## Types of changes

- [ ] Updated translation

## Checklist:

- [x] I have updated the documentation accordingly.


mRemoteV1/Resources/Language/Language.resx
```diff
@@ -2794,13 +2793,19 @@ Development Channel includes Alphas, Betas &amp; Release Candidates.</value>
...
   <data name="EncryptionTest" xml:space="preserve">
-      <value>Encryption Test</value>
+    <value>Encryption Test</value>
   </data>
   <data name="TestSettings" xml:space="preserve">
-      <value>Test Settings</value>
+    <value>Test Settings</value>
+  </data>
+  <data name="strPropertyDescriptionRdpProtocolVersion" xml:space="preserve">
+    <value>Sets the version of the RDP protocol to use when opening connections.</value>
```
farosch: Since the R ind RDP already stands for protocol it feels wierd to read 'RDP protocol'. I would change this sentence to 

> Sets the RDP version to use when opening connections

mRemoteV1/Resources/Language/Language.resx
```diff
@@ -2794,13 +2793,19 @@ Development Channel includes Alphas, Betas &amp; Release Candidates.</value>
...
+  </data>
+  <data name="strPropertyNameRdpProtocolVersion" xml:space="preserve">
+    <value>RDP Protocol Version</value>
```
farosch: same here

sparerd:
> I cannot open an RDP session when setting the protocol version to Rdc10.
The component check window does not show Rdc10 for me, either. I'm running Windows 10 1903 build 18362.175. It's the latest build and fully patched - why can't I use Rdc10?

I can't use it either - I thought it was just me. I can remove the Rdc10 option. In fact, is it even worth being able to select the RDP version? I wasn't sure about making it selectable, but figured someone will probably have some crazy use case where they need to use Rdp7 rather than Rdp9 (for example). If it just seems more complicated than it's worth, we can set it to always choose the highest version supported by the client.

> And unfortunately, the wallpaper is still being shown for me. Tried connecting to Windows Server 2016, 2019 and 2008R2. All were fully patched.

And you had `Display Wallpaper` turned off for the connection? Which RDP version did you use? Rdc6 doesn't support the property necessary to tell higher version to respect the performance flags.

ner00:
@sparerd I thested a compilation of this build (commit https://github.com/mRemoteNG/mRemoteNG/commit/3503652df33d1b1a8208af90b24a5adb2a0de6d0)

RDC10 setting also does not work for me on a Win10 1809 workstation, I receive the message: `Could not create RDP client. RDP protocol version Rdc10 is not supported on this machine. Please choose an older protocol version.` - it is totally supported though :)

As for the wallpaper, apart from RDC10, RDC7-9 (and 'Highest') were effective at hiding the wallpaper.

Additionally, when using RDC7+, it entails at least two additional performance settings which have been overlooked and are enabled by default:
_Menu and window animation_ (`disable menu anims`): This one can be checked by minimizing and maximizing a window, in RDC6 there is no gradual window restoration animation, it just pops-up/down crudely.

_Show window contents while dragging_ (`disable full window drag`): This is an old one, and easy to spot too, dragging any window should not drag the visual contents, only the frame borders until release.

sparerd:
> RDC10 setting also does not work for me on a Win10 1809 workstation, I receive the message: Could not create RDP client. RDP protocol version Rdc10 is not supported on this machine. Please choose an older protocol version. - it is totally supported though :)

When we try to create the class (`AxMsRdpClient10NotSafeForScripting`), it throws a `0x80040111 (CLASS_E_CLASSNOTAVAILABLE)` exception. I checked on my machine, and the COM class does exist (`{A0C63C30-F08D-4AB4-907C-34905D770C7D}`). The next most likely issue is that the class needs to be re-registered, but it seems very unlikely that all of us have a mis-registered class. The next possibility is that we don't have the system permissions to create that class (for some reason). Regardless, I think I'll just remove the Rdc10 option for now. We aren't utilizing any of its new functionality, so we really aren't losing anything.

> As for the wallpaper, apart from RDC10, RDC7-9 (and 'Highest') were effective at hiding the wallpaper.

Good to hear.

> Additionally, when using RDC7+, it entails at least two additional performance settings which have been overlooked and are enabled by default:

Since we haven't supported these before, this would be a new feature request to add these properties.

farosch:
Interestingly the about screen of the RDP clöient shows this:
image
So I should be able to use Rdc10.
I can't think of any use case where I would need to set the RDP version but like you said there might be some users that would need/want this feature and now that it's implemented I would like to keep it.

As for the wallpaper it does indeed work. I was just irretaded by the fact that Windows 2008R2 seem to display a blue background instead of a black one.

ner00:
> As for the wallpaper it does indeed work. I was just irretaded by the fact that Windows 2008R2 seem to display a blue background instead of a black one.

In any case, from personal testing, hiding the wallpaper already worked with Windows 2008R2(Std) prior to this new RDC implementation.

ner00:
> When we try to create the class (`AxMsRdpClient10NotSafeForScripting`), it throws a `0x80040111 (CLASS_E_CLASSNOTAVAILABLE)` exception. I checked on my machine, and the COM class does exist (`{A0C63C30-F08D-4AB4-907C-34905D770C7D}`). The next most likely issue is that the class needs to be re-registered, but it seems very unlikely that all of us have a mis-registered class. The next possibility is that we don't have the system permissions to create that class (for some reason). Regardless, I think I'll just remove the Rdc10 option for now. We aren't utilizing any of its new functionality, so we really aren't losing anything.

I've read a few things, which maybe you already came accross:
1. https://raw.githubusercontent.com/xenserver/xenadmin/master/XenAdmin/RDP/AxMSTSCLib.cs
2. https://docs.microsoft.com/en-us/windows/desktop/termserv/imsrdpclient10

1st link mentions the class `AxMsRdpClient10NotSafeForScripting`, not excalty sure in what context other  Citrix XenCenter;
2nd link mentions this: `Minimum supported client: Windows 10 [desktop apps only]` - that might be the reason why, it's possible that Win10 uses RDC9 for normal Remote Desktop Sessions?

As you said, no functionality is being lost by using RDC9.

sparerd:
I did find something interesting. When looking up the class `{A0C63C30-F08D-4AB4-907C-34905D770C7D}` (AxMsRdpClient10NotSafeForScripting) in the registry, I found this as a description: "Microsoft RDP Client Control - version 11". So if the `AxMsRdpClient10NotSafeForScripting` class is really for v11 of the protocol, then I guess we really don't have access to it yet as @ner00 is speculating.

Either way, I've removed the Rdc10 option for now. Thanks for the feedback so far, this has been helpful.

ner00:
@sparerd @farosch 
Any particular reason why Putty integration wouldn't work in a build from this branch? I just get a blank tab instead of the black terminal window tab.

sparerd:
@ner00 Not that I can think of. I'll check it out

sparerd:
@ner00 That ssh bug should be resolved now

PR #1594, opened 2019-09-26
Making a VNC connection to a host that is either off or does not have VNC causes the application not to respond for 20-30 seconds 

ryno2121:
Added code to report within 150ms a timeout error if no connection can be made on the VNC port for 150ms.

## Description
When attempting to make a VNC connection to a host that is off or does not have VNC installed will result in that the application will hang for 20-30 seconds, because it attempts to connect to the host for 20-30 seconds before continuing with the program. I have used a call back method with a test connection to test the connection to the host using the user configured VNC port, and if the command runs for more than 150ms it will through a connection timeout error. 

## Motivation and Context
It has been annoying for us when attempting to connect to a host that is off and then the application hangs for 20-30 seconds before we can continue with a next connection. The code with the call back method I have added solves this, making the application more pleasant for us to use. 

## How Has This Been Tested?
In our environment we use mRemoteNG for all our VNC remote connection. mRemoteNG gets used 24 hours a day to make remote VNC connections to various hosts. With the new code applied I tested it with all my colleagues for 2 days with no issue reported. 

## Types of changes
The code was only added to the Connection.Protocol.VNC.cs file. 

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.


farosch:
@ryno2121 I think something went wrong when you tried to create this PR. You are just trying to merge the `develop` branch into the `ssh_focus` branch which does not make any sense at all. 🤔

ryno2121:
My apologies farosch. I will do so.

farosch:
No worries! :)
Your effort to improve mremoteng is greatly appreciated! Just ask if you need any help!

Issue #1071, opened 2018-08-11
CefSharp Browser engine


spisarau:
Could you add Chromium Engine to mRemoteNG?

https://github.com/cefsharp/CefSharp

leaskovski:
@sparerd I have just started knocking something up for this.  Whilst it looks pretty simple to add in, for some reason I am having difficulties with getting the actual control to render in the application.  Any pointers for my current implementation?

amaitland:
> I have just started knocking something up for this

@leaskovski Push your changes to `GitHub` and post a link, I'll have a quick look and see if anything stands out.

leaskovski:
@amaitland you can find my fork of the repo here, https://github.com/leaskovski/mRemoteNG

amaitland:
Does google.com load? I don't see anything glaring from a CefSharp point of view. 

The finalizer you've added to Dispose of the browser is almost guaranteed to cause problems,  hopefully whoever reviews your PR provides a constructive review.

Finalizer should only free unmanaged resources,  see https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-disposeboolean-overload

leaskovski:
@amaitland , ok thanks, I will review that dispose doc.  The google site does load as we get a title change event trigger fired and thus to page title then gets loaded into the connection tab caption, so that looks to be working ok.
I am going to create a sample app from scratch and see what is different between the two.  I have a feeling it might be related to how the web browser component gets attached to the tab.

amaitland:
You can use https://github.com/cefsharp/CefSharp.MinimalExample for testing purposes.

leaskovski:
@amaitland just realised that you also work on CEFSharp!  Doh.  Anyway, I got around my issue.  Had to change how the browser control was being added to the control collection and this solved the loading problem.  So now have a kind of working version of the implementation.  Might submit a pull request.

spisarau:
Recently Microsoft released a new version of Microsoft Edge browser. So it can be used as external tool and integrated into mRemoteNG window without any problem. 
The following startup arguments allow to start ech host in a separate tab
`%Hostname% --start-fullscreen --user-data-dir=C:/TEMP/%Hostname%`

leaskovski:
I dont know why I didnt even think of doing that!!!  lol....
`chrome.exe --app=%HostName%`

PR #1690, opened 2020-02-11
Replace GeckoFX with CefSharp

leaskovski:
Added the CEFSharp component to the project. ~~This is a Web rendering engine that is based on Chromium meaning that as well as being able to use IE and Gecko (Firefox), users will now also have the option to use CEF (Chromium) as their rendering engine.~~ This is a Web rendering engine that is based on Chromium meaning users will be able that as well as being able to use IE and Chromium as their rendering engines.
This will mean the use of Chrome Remote Desktop as a URL can be used to provide connections to machines that use this protocol.

## Description
Added a new option to "Rendering Engine" called "CEF (Chromium)".  This requires CEF Sharp as being added to the project via nuget, and add a chromium based rendering engine to the application.
This covers the feature request #1071

## Motivation and Context
Chrome Remote Desktop is a protocol provided by Goolge to allow remote connections to machine.  This can be accessed via a website (https://remotedesktop.google.com/access), however this site (being Google) doesn't play well with other browsers, so neither the IE or Gecko engines work.  By adding CEF, the site renders fine and allows connections to machines hosted under this tool.

## How Has This Been Tested?
Basic testing has been completed by using the new engine for the original mentioned use case of Chrome Remote Desktop.
Tested under Windows 10 and Windows 7 environments.
Tested using Portable build only.
Tested persistence of logins to to site mentioned above.

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation, if necessary.


farosch:
First of all thank you very much @leaskovski for this PR! 

@sparerd @kmscode 
I would much more prefer to implement chromium **exclusively** into mremoteNG. What do you guys think about this?
-The IE based web browser control is nightmarish wehen it comes to security vulnerabilities. Even Microsoft itself does not recommend using IE anymore
-Chromium has the highest compatibility
-Chromium uses the Microsoft Certificate Store which makes it easier for system administrators to handle trusted certificates, especially when they come from a Windows Certificate Authority
-There is no real benefit (at least to my knowledge) that comes with choosing between different HTML rendering engines
-Simplification of code

farosch:
I have decided to close #1761 because IE support might still be needed by users to maintain legacy hardware. There's absolutely no reason to have both GeckoFX and CefSHarp build into mRemoteNG  though - There is no real benefit to it and shipping with both rendering engines would blow up the size of mrng.

farosch:
@leaskovski I've added a download handler to handle file downloads in cef and also added a link handler, so that I could use cef for the help window also.
This allowed me to completely replace gecko with cef.
Does this look good to you? Any suggestions for improvement?

leaskovski:
@farosch just looking over the code.  Tried to run it on my dev system and it chucks an error because of the help files not being located.  Is help excluded for the portable versions of the software?  Might be worth putting an file space check around that loading?

farosch:
@leaskovski I think the Firefox folder you still see in your output folder is left over from previous builds.
Use _Build_ > _Clean Solution_ from within VS before building this branch.
I don't have a Firefox folder in any build and Accessing the help files works just fine for me in every build, too.
I think you get that error simply because you haven't set up your environment properly and the help files don't get build on your system.
Check this page out: 
https://github.com/mRemoteNG/mRemoteNG/wiki/Development-Environment

farosch:
@leaskovski
Whenever someone deletes the _Help_ folder from within the mrng directory they will be greeted with a runtime error upon opening mrng.
I'm pretty sure I could program a check around that and disable the help button when the directory cannot be found on startup.
On the other hand users could delete any other file from the directory also, which would result in some other error that is not handled.
I'd just go with the statement that everyone that deletes folders from a programs' directory will have to live with the consequences that come with it.
I don't delete my system32 folder and bame microsoft for the results, either...

leaskovski:
> @leaskovski
> Whenever someone deletes the _Help_ folder from within the mrng directory they will be greeted with a runtime error upon opening mrng.
> I'm pretty sure I could program a check around that and disable the help button when the directory cannot be found on startup.
> On the other hand users could delete any other file from the directory also, which would result in some other error that is not handled.
> I'd just go with the statement that everyone that deletes folders from a programs' directory will have to live with the consequences that come with it.

I totally understand your sentiment on this.  If users don't have complete distributions, then they should expect errors.  The only thing is, I just downloaded the portable package from https://mremoteng.org/download, and that doesn't contain a "Help" folder, hence why i was wondering if the build process for the portable builds needs updating?

farosch:
You're right, the prerelease doen't contain the Help files.
This does not have anything to do with this PR though, the last prerelease is from Mon, Sep 02, 2019.

Issue #181, opened 2016-10-19
Replace PuTTY with SSH.NET


kmscode:
Replace PuTTY with SSH.NET

klonos:
Just curious, what are the major advantages of SSH.NET over PuTTY?

jumoog:
no idea.. but maybe better Linux support?

sparerd:
There are a few reasons we've been looking into replacing PuTTY as the default ssh/etc client:
- Tighter integration. We currently spawn a new PuTTY process off of mRemoteNG. It's essentially acting like any other external program you can define, just with a little more built-in support.
- Better security. Because PuTTY is called as an external program, we have to pass host/username/**and password** data to it in clear text over the command line.
- Better customization. Since SSH.Net is a library, we can control every minute aspect of the connection that it establishes and how it is displayed to the user.
- As @jumoog said, this also allows for better Linux support. SSH.Net is written in .Net Core which runs on Linux. Removing the PuTTY requirement is a good step toward cross platform support.

For those that still like PuTTY, we will be keeping it around as an option for handling ssh/telnet/RAW/etc. However, it will not be the default for new installations.

sparerd:
Yea, we could still load and use putty session info from the registry. It probably wouldnt be the primary way of saving ssh connection records though

klonos:
@sparerd thanx for taking the time to explain. All these seems pretty solid reasons as to why the change. It's just that with the change, I'm not sure whether certain features that I'm used to from PuTTY would still be available (silly things like the selection auto-copy and the right-click paste). That's all.

jumoog:
auto-copy and the right-click paste is priceless

igor2x:
Just my two cents. The following bug on JIRA was pointed to this Github bug:
https://mremoteng.atlassian.net/browse/MR-883

sparerd:
This is all great feedback. We wont be replacing putty right away, there
will be plenty of time to make sure the native ssh client has the right
features

mcarbonneaux:
must be compatible with peagent protocol for key forwarding, and then can be used with pageant or alternative like keepass with keeagent....

kmscode:
@mcarbonneaux 

Thanks for the additional info. This is largely just a place holder at this time. We don't have any solid time frame as to when this could be implemented and due to various other factors, its unlikely that PuTTYNG will be removed (it just wouldn't be the default most likely).

pedro2555:
[...]

PR #1709, opened 2020-03-11
Feature/serial UI integration

azet:
Add Quick Integration to use Serial Connections (e.g. for USB Serial/COM devices on Windows) within mRemoteNG for example to Debug Network devices via Serial Ports or other Serial Protocol connected devices.

## How Has This Been Tested?

Was been tested in Visual Studio 2019 with the newest available .NET Framework

image

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 ### Added
+- Added support for Serial connections in the Connection Drop-Down menu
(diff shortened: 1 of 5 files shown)


sparerd:
Thanks for the PR @azet! Unfortunately, there is a build error. You will need to add a language entry for `strSerial`.

azet:
@sparerd: I added that to Lang as best as I understand it. For some reason this built just fine for me without adding that.

farosch:
Serial connections would have to be fully implemented into mremoteng as its own connection protocol for this PR to be acceptable.
also further relying on PuTTY should be avoided due to #181.

Issue #1713, opened 2020-03-17
Sound redirection does not work if Clipboard redirection is set to No


GurliGebis:
## Expected Behavior

Sound redirection should work even with Clipboard redirection turned off.

## Current Behavior

When Clipboard redirection is set to No, Sound redirection does not work - no sound is passed through, and there is just silence.
Trying to playback audio on the remote machine just makes the playing application hang.

## Possible Solution

Enable Clipboard redirection as a workaround.

## Steps to Reproduce (for bugs)
I have run a git bisect, and found the commit that caused the problem, I can however not spot what is wrong with the code.
Here is the output of the bisect:

<pre><code>
$ git bisect log
git bisect start
# good: [068f5942e73a0b4e3c461d861216788ac5a4958a] updated changelog and bumped version to 1.76.20
git bisect good 068f5942e73a0b4e3c461d861216788ac5a4958a
# bad: [3e354868ea0499ad89880fc240313e66e513d72e] also fix installer project cert password reference
git bisect bad 3e354868ea0499ad89880fc240313e66e513d72e
# bad: [20f3b19f2991377bec5e8e156a37543e049b1610] Restored reconnect all, correct tab collection now used instaed of custom one
git bisect bad 20f3b19f2991377bec5e8e156a37543e049b1610
# bad: [d307adbbefbf011b8d01e8311d198e0f76de30b6] Merge pull request #1228 from farosch/external_links
git bisect bad d307adbbefbf011b8d01e8311d198e0f76de30b6
# good: [b223b121d1c9c47b03d9e155e6ee54724dd9edc2] minor visual fix
git bisect good b223b121d1c9c47b03d9e155e6ee54724dd9edc2
# bad: [78ef0fdaf48920511874ebfc5c9bce85a6bbf80d] Merge pull request #1107 from mRemoteNG/321_hidpi
git bisect bad 78ef0fdaf48920511874ebfc5c9bce85a6bbf80d
# bad: [6b280b5aa44275a0bde0e058aa78a09654bc5d97] Merge pull request #1145 from st-schuler/reconnect-patch
git bisect bad 6b280b5aa44275a0bde0e058aa78a09654bc5d97
# bad: [fbd040786350a2de9354af38c3d7c6799895ad9c] added redirect clipboard to xml schema for v2.7 confcons
git bisect bad fbd040786350a2de9354af38c3d7c6799895ad9c
[...]

Issue #1632, opened 2019-10-27
1.77.1 breaks RDP drive and sound redirection


klui2k1:
## Expected Behavior

Host drives should show up and there should be sound while using RDP. Microsoft's msts works.

## Current Behavior

No host drives show up when opening Windows Explorer, under Windows 7 or 10.

## Steps to Reproduce (for bugs)

1. Define an RDP connection
2. Under config, ensure Protocol is set to RDP, Version is set to Rdc6, Rdc8, or Highest. Probably doesn't matter.
3. Under config, ensure Redirect Disk Drives: Yes, Sounds: Bring to this Computer.
4. Connect then play a sound and open Windows Explorer
5. Sound doesn't play; Windows Explorer doesn't show host drive.

## Your Environment

* Version used: 1.77.1
* Operating System and version (e.g. Windows 10 1709 x64): Windows 7 x64 - up-to-date

GurliGebis:
@klui2k1 are you seeing this problem with version 1.77.0 ?
If not, that helps narrow down where the problem was introduced :)

klui2k1:
It also happens under 1.77.0.

GurliGebis:
Okay, but it was introduced in the 1.77 branch? (not in the latest 1.76) 🙂

klui2k1:
Correct, introduced in the 1.77 branch, not in 1.76.

It could be a regression issue with allowing which drives to map per https://github.com/mRemoteNG/mRemoteNG/issues/317

GurliGebis:
Nope, it is caused by Clipboard redirection being set to No, which causes it to fail for some reason :(

klui2k1:
Maybe related to a change in behavior in MS's main branch of RDP described at https://social.technet.microsoft.com/Forums/en-US/04ff5091-491d-4d80-ba0b-1819959904de/rds-2012-r2-disable-local-drive-redirection-with-working-clipboard?forum=winserverTS

GurliGebis:
It is only blocled until it has been reviewed by someone with write access to the github project

xbliss:
> Nope, it is caused by Clipboard redirection being set to No, which causes it to fail for some reason :(

Facing this again here: https://github.com/mRemoteNG/mRemoteNG/issues/1879 

Is there a work around or fix?

Issue #1465, opened 2019-05-28
REGRESSION: Smart Cards redirection to Remote Desktop not working


bluikko:
In 1.77.0.41252 the logon to Microsoft Windows servers with Smart Card does not work anymore.

It used to work a few versions ago.

## Expected Behavior
If "Smart Cards" is set to "Yes" (either directly or inherited), then opening a Remote Desktop session to a Windows Server should allow Smart Card logon.

## Current Behavior
Logon with Smart Card is impossible, user is presented with a logon screen.

Running "mstsc.exe" outside of mRemoteNG and Smart Card logon works to the exact same servers, as expected, when Smart Card redirection is enabled. The only difference that I can see is that running "mstsc.exe" manually works, whereas connection from mRemoteNG shows no available Smart Cards in the logon screen (redirection not working?).

Usually (and in earlier versions), below "Steps to Reproduce" ended in connection being opened after step 3 and user presented with desktop of the remote machine.

## Possible Solution
Suspect redirection of Smart Cards is not working since the logon screen shows no available smart cards when clicking on the "smart card logon option" icon.

## Steps to Reproduce (for bugs)
1. Enable "Smart Cards" for Remote Desktop connection in connection options section "Redirect".
2. Connect to a Windows Server by opening the connection.
3. When prompted for credentials (using Network Level Authentication/NLA, not yet in Windows logon screen), choose Smart Card and type PIN and username hint.
4. Connection is opened, but Windows Server logon screen is displayed, no Smart Cards are available at the logon screen.

## Your Environment
* Version used: 1.77.0.41252
* Operating System and version (e.g. Windows 10 1709 x64): Windows 10 1803 x64

sparerd:
@bluikko Is there anything in the log file? (https://mremoteng.readthedocs.io/en/latest/troubleshooting.html#logfile)

We haven't changed anything around the smart card redirect feature. We just pass the true/false value to the RDP client. Something may be failing and preventing that value assignment from happening. The log should hopefully have that info

bluikko:
It looks like all sharing is not working. Clipboard does not work either to Remote Desktop.

Will try to log soon.

bluikko:
I was mistaken in the message above - clipboard does work.

And after I checked the settings by disabling inheritance + enable smart card and clipboard in host settings, now smart card seems to work also?

I was sure that I especially tested this by disabling inheritance and set "Yes" to the individual host settings.

[...]

PR #951, opened 2018-04-15
Clipboard sharing setting implementation for RDP protocol

CrAbelleira:
## Description
Implementation of **clipboard sharing setting for RDP protocol**.

## Motivation and Context
This implementation provides de posibility of enabling or disabling clipboard sharing on RPD connections which can be usefull in multiple scenarios.
It solves the issue #416 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.

PR #1714, opened 2020-03-17
Corrected how clipboard redirection is done

GurliGebis:
## Description

Previously DisableRdpdr was used, which broke smartcard and sound redirection.
By changing it to use AdvancedSettings6.RedirectClipboard, it works without breaking other stuff.

## Motivation and Context

The problem was introduced in this PR: #951
This change fixes the following issues: #1465, #1632, #1713

## How Has This Been Tested?

I have tested with the 1.77.1 build without the change, and if clipboard redirection is set to No, drives and sound is broken - setting it to Yes and they work.
With this change they work with the Clipboard setting set to both Yes and No.
I have also tested that the clipboard setting is working, setting it to No results in the clipboard not being carried across, setting it to Yes, and it does.

## Types of changes

Bug fix

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - #1589: Update SQL tables instead of rewriting them
 - #1337: Unhandled exception after closing mRemoteNG
 - #359: Making a VNC connection to an unreachable host causes the application to not respond for 20-30 seconds
+- #1465: REGRESSION: Smart Cards redirection to Remote Desktop not working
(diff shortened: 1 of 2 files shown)


GurliGebis:
For more info:
Here is what was being used:
https://docs.microsoft.com/en-us/windows/win32/termserv/imstscadvancedsettings-disablerdpdr (This one disabled way too many things)

Here is what should be used instead:
https://docs.microsoft.com/en-us/windows/win32/termserv/imsrdpclientadvancedsettings5-redirectclipboard (This one only changes the clipboard setting)

farosch:
Great find! Thank you very much! :)

Issue #1688, opened 2020-02-06
Integrated Browser / Menu / Bookmarks / URL Link


thorntonp72:
Hi,

Firstly. Awesome tool and great work.

When I use the browser integration for either Firefox or Internet Explorer, there is never an option to either show/enable either the Menu, Bookmark or URL options at the top.

## Expected Behavior
One would expect to be able to either use the setting defined by the Browser or be able to show/hide these options. At the moment it only opens the browser into a full screen mode with no menu options or bookmarks at all. The reason for wanting this is that the Web connection does not allow for compilicated URL connections such as  http://myurl.com:7001/console. These have to be setup using the External Tool, but this then opens the window in a new window and not inside mRemoteNG.

## Current Behavior
There is no option to enable/disable or view these options.

## Possible Solution
1. Allow some options to not go into full screen mode and/or have these as passable options.
2. Allow more options for the URL ( such as http://my.host:port/my/url ) for the existing Web integration.

## Steps to Reproduce (for bugs)
1. Configure a new Browser Connection 
2. Open it
3. No menu/bookmark or URL's appear

## Context
To be able to integrate this into the mRemoteNG window with more options would be ideal. Currently I can do this by using the External Tools option but it open a window outside of mRemoteNG which is not what I would like.

I can't also easily integrate this into the Web Connection as I need to pass full URL's including port numbers and paths. i.e. http://localhost:7000/my/path (This does not work), If it did then this would also fix the problem.

## Your Environment
* Version used: 1.76.20.24615
* Operating System and version (e.g. Windows 10 1709 x64): Windows 10 x64

Thank you,

Paul

tschroeder13:
Hey all, I fully agree to thorntonp72 and would appreciate this feature!

robertpopa22:
**Fix available for testing**

**Commit:** [`99445518`](https://github.com/robertpopa22/mRemoteNG/commit/99445518fcc8542dbdf558d516e1a2cdf2712053)
**Branch:** `main`
**What changed:** Add menu items for bookmarking the current page and managing

**Download latest beta:** [v1.81.0-beta.2](https://github.com/robertpopa22/mRemoteNG/releases)

Please test and report if this resolves your issue.

---
_Automated by mRemoteNG Issue Intelligence System_

Issue #1614, opened 2019-10-07
Open link in http connection creates ghost window for external link


nmat:
A ghost window is created when trying to access a link that goes outside of the mremoteng window. This is really a ugly way to handle external links and so should be fixed or opened in the default browser.

## Expected Behavior
Either get a question to open in default browser or follow the link in a new tab.

## Current Behavior
Creates a ghost window and nothing happens.

## Steps to Reproduce (for bugs)

1. Login to the administrative interface for a synology diskstation
2. Go to Control Panel -> Update and Restore
3. Try to go to version information for installed DSN software

BOOM creates a ghost window and nothing happens.

farosch:
dev note:
looks like there is no handling of external links in http/https connections. This happens for connections using gecko as rendering engine. I will need to verify if this is also the case when using the IE engine.

farosch:
No longer relevant since merging #1690

Issue #1460, opened 2019-05-24
Windows Admin Center not usable in a http/https connection


Songxwn:
Windows Admin Center is not usable in a http/https connection.
There are rendering and probably also functional issues calling the windows admin center from within mRemoteNG with the gecko engine.
Using the IE engine a popup is shown to use a different browser.

nmat:
Not sure what you mean here. Can you please provide more information on how to reproduce this and what you are doing?

farosch:
This is imho caused by the version of the gecko engine we currently use in mrng.
The Windows Admin Center works perfectly in newer versions of firefox.

farosch:
#1477 would make this possible, too

farosch:
It kinda works in GeckoFX but not as nicely as with CefSharp

Issue #964, opened 2018-04-24
Gecko Rendering no longer working on Windows 2012 R2 Host


goggles-mcduff:
Connections to WEB (HTTP or HTTPS) services using the Gecko rendering no longer work on a Windows 2012 R2 host using version 1.76.3.32942.

Having recently updated to 1.76.3.32942 all my connections for web services using Gecko rendering now fail, a blank window is presented, 

Connections using Internet Explorer rendering still work, however as this does not support connections where a self-signed certificate is used on HTTPS connections, this does not help!

If I uninstall and reinstall 1.76.0.25712 then everything works again.

I have also tried 1.76.0.36642, and this also does not work.

I have tried this on two different 2012 R2 Machines, and both exhibit the same problem.

The same version installed on Windows 10 Desktop works fine.

PR #1761, opened 2020-05-23
Implemented CefSharp as rendering engine exclusively

farosch:
**This PR is based on #1690 by @leaskovski** 

## Description
This PR removes geckoFX and IE as rendering engines for HTTP/S connection ans implements CefSharp as the default rendering engine. This also affects the in-app help files created with sphinx.

The AboutWIndow has been stripped down to a simply window showing basic information and web links to read credits, changelog and license instead of embedding them into the application.
This looks much cleaner and removes a lot of code.

## Motivation and Context
- The IE based web browser control is nightmarish wehen it comes to security vulnerabilities. Even Microsoft itself does not recommend using IE anymore
- Chromium has the highest compatibility
- Chromium uses the Microsoft Certificate Store which makes it easier for system administrators to handle trusted certificates, especially when they come from a Windows Certificate Authority
- There is no real benefit (at least to my knowledge) that comes with choosing between different HTML rendering engines
- Simplification of code
- Less code to maintain

## What problem does it solve?
#516 #1071 #964 #1688 #1614 #332 #1460

image

image

## Types of changes

## Checklist:


farosch:
@nmat pointed out, that some people might still need IE rendering to be able to connect to older hardware.
I don't know if there are enough people affected by this for us to justify maintaining IE support or if we could get away with these users having to add an external app for IE.

leaskovski:
@farosch there are some sites that dont play nicely with chrome, specifically things like older versions of sharepoint, vsphere and SCVMM.  If an external app using IE works and integrates fine, then it shouldn't be an issue.  It should be worth checking this as part of any validation, and potentially putting the config for IE as an external app into the documentation.  Having it run in integrated mode would be ideal as then it would seem like it is running as before.

Issue #1202, opened 2018-12-06
Remove Citirix Support


kmscode:
Citrix Support will be removed  in 1.78

reasons: Simpler build environment, stepping stone for x64 builds, lack of test environment, lack of interest based on surveys

kmscode:
Installing the Citrix Receiver in Appveyor takes between ~1-3+ minutes for EACH configuration (Release, Release Portable, Release Installer) for every build request. This adds anywhere between 2 and 10 minutes for a build to complete in Appveyor.

PR #1763, opened 2020-05-24
Remove ICA (Citrix) support

farosch:
## Description
This PR completely removes citrix support from mremoteNG

## Motivation and Context
https://github.com/mRemoteNG/mRemoteNG/issues/1202
I know this has originally been scheduled for 1.7**8** but I'd rather like to do this asap.
Appveyor hasn't replaced our jenkins build server yet and before having to configure appveyor twice I'd like to get rid of Citrix now so when we implement the devlivery process through appveyor we can instantly provide 64bit builds without having to reconfigure appveyor again as soon as citrix is being removed.

## Types of changes

 - Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


kmscode:
I'm totally OK with this in general.

I believe that we sent out a survey and ICA wasn't popular at all... I'll see if I can dig out the details just to re-confirm.

@farosch - I didn't review the code yet, but off the top of my head I would suggest that any of the ICA related properties within the connections xml should simply be ignored when loaded and removed on save.

This may require a new confcons schema...

farosch:
@kmscode Thats exactly how it works. All properties that don't exist anymore are ignored and connections that used to use the ICA protocol are converted back to the default protocol, RDP in this case

lukeattard:
I say we remove ICA support an maybe have it as a plugin for the rare case someone needs it

PR #1817, opened 2020-07-15
Update troubleshooting.rst

kevinduongcalabrio:
## Description
Backslash needed to be escaped in doc


farosch:
> Backslash needed to be escaped in doc
Very true, thanks for this!

Issue #1427, opened 2019-04-30
RDP doesn't respect scaling


luisarnauth:
After upgrading to 1.77.0.41307 the RDP connections don't respect the desktop scaling , showing the remote desktop in 100% scaling.

Prior to this upgrade, the remote desktop connections inherit the desktop scaling.

2ramiro:
Got the same problem.
Also an option to manually scale the size would be nice for hosts that doesn't support scalling for exampe 125% , 150%  , 200 % .

warmdev:
Same problem. Also like @2ramiro suggested, would like to see manual scaling or zooming options when scaling is not properly detected.

7OH:
With 2 monitors with different scale on each - situation is more "fun" (#1438)

mdio33:
Same problem here... kind hard if you have a 4k screen.

https://i.postimg.cc/5t3cNp6G/mremoteng.jpg

But the worst is that tryed to revert to v1.76.20 and lost all saved credentials :(

kirsan31:
it's very strange. But this is not working for me even in v1.76.20 :( I connect from Win7 machine with 125% dpi to various Win10 machines and all of them showing in 100% dpi :(
I began use mRemoteNG exactly because this feature, and now it's suddenly stop working :(( I use portable version, was trying 1.77.1 and clean run - no luck :( Any advise?

And I agree with all above that, manually selecting dpi per connection very needed. Native Win10Store app and JumpRDP (MacOS) have this feature...

luigirosa:
Same problem here, a fix would be very helpful.

tonibony:
A workaround for this issue on Windows 10: Find `mRemoteNG.exe` file, open "Properties" -> "Compatibility" -> "Change high DPI settings" -> "High DPI scaling override", turn on "Override high DPI scaling behavior" option and set "Scaling performed by:" to "System" or "System (Enhanced)", not "Application".

mdio33:
> 
> 
> A workaround for this issue on Windows 10: Find `mRemoteNG.exe` file, open "Properties" -> "Compatibility" -> "Change high DPI settings" -> "High DPI scaling override", turn on "Override high DPI scaling behavior" option and set "Scaling performed by:" to "System" or "System (Enhanced)", not "Application".

Thank you very much Toni, that soved my problem.

;)

Hecatron:
[...]

PR #1898, opened 2020-11-15
Fix RDP local desktop scale not taking effect on remote

jobarner:
## Description
RDP remote desktop scale was calculated incorrectly, resulting in local 125%/150% translating to remote 100%

## Motivation and Context
When working with a high-DPI display with a scaling factor, the remote machines are set to 100% due to this bug. This results in RDP session where everything is too small.
#1427 RDP doesn't respect desktop scaling

## How Has This Been Tested?
Tested on my 125%-scale setup: laptop monitor of 1920*1080 and + 4K external monitor. My change works great on both monitors.
This change is a one-liner, only in the RDP connection, so I believe it's low-risk.

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Updated translation

## Checklist:

- [ x ] I have read the **CONTRIBUTING** document.
- [ x ] My code follows the code style of this project.
- [ x ] All Tests within VisualStudio are passing
- [ ] This pull request does not target the master branch.
- [ x ] I have updated the changelog file accordingly, if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 ### Added
+- #1427: Fix RDP local desktop scale not taking effect on remote
 - #1770: Added missing RDP performance settings
 - #1332: Added option to hide menu strip container
 - #545: Option to minimize to system tray on closing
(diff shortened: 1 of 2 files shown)


jungleboydotca:
Made this change into 1.77.1 and successfully built.

Thank you! I can finally use (mostly) properly scaled RDP sessions; unfortunately higher RDP versions seem to mess-up font rendering, but I'm fine using v6 for now.

tecxx:
hello, could you check my here comment https://github.com/mRemoteNG/mRemoteNG/issues/321#issuecomment-777828387
i think more code is needed, if i understand the ms docs correctly, the scaling should be fixed values like 140, 180, etc (i may be wrong though.)

PR #2011, opened 2021-08-01
removed screenshot manager

farosch:
## Description
Removed the screenshot manager from mRemoteNG.

## Motivation and Context
First of all I don't think a screenshot manager is something mRemoteNG should offer. There are a lot of screenshot managers (e.g. Greenshot, ShareX) that have a ton of features for anyone who really needs a professional screenshotting tool. The implementation in mRemoteNG is only rudimental and does not really serve a purpose, as modern versions of Windows itself features better tools (`Win` + `Shift` + `S` / `Win` + `V`). 
As this feature only possibly introduces bugs and takes time away from maintaining the main features of mRemoteNG I think we should get rid of it.

## How Has This Been Tested?
No testing necessary.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Updated translation

## Checklist:


Kvarkas:
Perhaps its ok to remove it

simonai1254:
I don't use it a lot as well, though I can think a quick button to store just the connection screen on disk on a predefined location for easy documentation would be nice.
This just because removing this option results in the need to take a screenshot or with the connection panel included or with manual selection of the connection frame only.

Though as I don't know how hard it is to maintain I guess this can be removed and thought of as an additional feature to reimplement in the future.

farosch:
IMHO this is something that should be either handled by OS defaults (e.g. `Win+SHift+S`) or another third party program as anyone having this need probably has it for other tools, too.

simonai1254:
Using a dedicated screenshot utility is a viable option, though I am a fan of having a button that directly takes a screenshot of the correct area and not having to draw the special area every time (the connection panel is useless). 

E.g. I use the built in feature of Zoom to create a screenshot of the shared screen (2 clicks, are selected correctly, file saved and folder containing file opened automatically) because it's easier than opening a screenshot utility (one key combination to open, a finnicky selection of the screen area that I might need to redraw and miss the content displayed at that moment, then manually saving the screenshot to disk).

I would not mind having that feature implemented into mRemoteNG similar to Zoom as there are times that I would like to document my steps directly on the mRemoteNG device using screenshots.

As I already said if this feature is hard to maintain I definitively can go without, I am capable of doing the screenshots with dedicated utilities. It would just be for improved usability, the way it's implemented right now with a special screenshot manager is definitively overkill.

PR #2063, opened 2021-11-04
Thycotic Secret Server Interface

tecxx:
## Description
adds functionality to fetch credentials from thycotic secret server via it's API (both username/password and single sign on is supported)

## Motivation and Context
prevent credential theft - storing secrets in local config files is always a bad idea. since TSS supports access to secrets via it's API it would be nice if mRemoteNG could interface to it

## How Has This Been Tested?
using it in production on a daily basis in our company for nearly a year

usage is simple. instead of specifying username and password, specify the secret ID only (format: SSAPI:xxxx)
image

a popup will ask for API data (only once)
image

the plugin will then fetch the credentials and provide them to the connection. no secrets are stored in memory.
currently this is implemented for RDP and SSH.

## Types of changes
- [X] New feature (non-breaking change which adds functionality)
api access code is segregated in it's own project (DLL), and does not modify any mremote code. i chose this approach to be as modular as possible. the code can be merged to the main exe instead, if required. but maybe other credential storage solutions will be added later, so keeping this seperate is a good idea?

shortly before initiating the connection the DLL is called to fetch the credentials. for this two minor changes have been made to RdpProtocol6.cs and PuttyBase.cs. if there is a better way to do this, let me know, but afaik there is no "plugin system" for external credential providers, correct?

## Checklist:


kramttocs:
This integration is excellent to see. Thank you @tecxx .

I looked into it yesterday but am getting:

----------
A connection protocol error occurred. Host: "REDACTED"; Error code: "0"; Error Description: "Secret Server Interface Error: authentication error: Login failed."
----------

If I visit:
REDACTED.secretservercloud.com/SecretServer/winauthwebservices/SSWinAuthWebService.asmx
I get:
An error has occurred in Secret Server Cloud.

Any suggestions?

Kvarkas:
@kramttocs are you sure credentials are correct?

kramttocs:
@Kvarkas  This is using SSO (Microsoft ADFS). I would try using the local username/password for Thycotic but I do not have access to that password.

tecxx:
Hello,
i am sorry but in that constellation (TSS Cloud AND SSO) i am unable to assist. i only have the on-premise installation in my environment, and only one customer with the cloud version. i could ask for an account there to debug it, but they do not use ADFS/SSO so that will not help me very much...

tecxx:
@kramttocs i just tested this for a customer who uses secret server cloud (no sso, but with OTP). it works fine. maybe try the latest nightly build. we made a few improvements to the connector, which might solve your isse.

kramttocs:
Thanks @tecxx . I'll check it out.

kramttocs:
Just installed the latest nightly and was going to test this out again. I see there is an API User ID field now. I entered SSAPI:170 into it but when trying to connect I get an error of:

----------
WarningMsg
05/02/2022 10:16:24
A connection protocol error occurred. Host: "REDACTED"; Error code: "0"; Error Description: "Secret Server Interface Error: Input string was not in a correct format."
----------

tecxx:
Use the ID only (170)
The prefix was used before we added the dedicated fields.

kramttocs:
Thanks @tecxx . Afraid there is still some information I don't have that would require assistance from the internal security team.

I do see in Thycotic User Preferences that there is an API Token option. Is that something that could be used for this to avoid the SSO config?

tecxx:
hy,
the answer is always "it depends".

if your thycotic secret server is set up to allow windows integrated authentication (also known as SSO or pass-through-authentication) you can tick the SSO checkbox and the connector will try to talk to the API with the credentials you are currently logged on. in this case username and password fields are greyed out, naturally.

if SSO is not set up (and that will be most installations) you have to specify username and password to talk to the API. in this case do not tick the SSO checkbox.


now if your secret server admin has protected the API with a second factor, you have to specify the OTP code as well. which code this is depends on how your TSS is set up.

in our case we have protected TSS with Duo Security. so we specify username, password, and the 6-digit OTP code from the Duo Security App.

if you don't use Duo Security you can use the builtin OTP function (as you mentioned). in this case you first have to register your user in TSS with your 2-factor-app.

do note that when writing the connector i didn't test all the possible combinations that TSS offers for authentication. the one we (and our customers) mostly use is duo, so thats the one we made sure works correctly :)

let me know if that helped

speraltamontoya:
Hi!

I am trying to make the secret server api configuration pop up appear, but it does not appear after entering the value SSAPI:42 in "API User ID" I have also tried it in Username .
I have mremoteng version 1.77.2-nb (1746af9) installed

How do I make the popup appear?
Can it be configured directly in the configuration file?

thanks, regards.

tecxx:
hy, not sure which version/build exactly but i enhanced the feature to potentially add other external credential vaults as well in the future, so the config would look like this now:

image

do you see the "external credential provider" dropdown in the connection options already? if not, can you build the latest nightly yourself?

i should replace the screenshot in the docs at some point (https://mremoteng.readthedocs.io/en/v1.77.3-dev/howtos/credvault.html)

Kvarkas:
I am thinking perhaps we could make it as a plugin-ish so then we could remove it from core, that allow users to have simple interface then they dont use it, and if they need such just add a plugin! :) Also authority for plagin will be yours so you could do a release of new versions any time without dependency on main app release cicle :) just thinking loudly...

PR #2084, opened 2021-12-14
fix some more inconsistencies with "StartProgramWorkDir"

tecxx:
tried to fix a few more of the test cases and noticed that the property "StartProgramWorkDir" is not consistent with "StartProgram". made a few small fixes and some more test cases are now ok.


farosch:
@tecxx Looking good! Thanks a lot for this.

PR #2115, opened 2022-01-11
Fix tests that were broken by PR #2100

Vest:
## Description
Every time when a new property is added to a connection node, the tests **ConfigWindowGeneralTests** should be corrected. The code there checks what properties **must** be available for different connection types, and what not.

Currently PR #2100 broke the tests that were green.

## Motivation and Context
Make the project better

## How Has This Been Tested?
VS 2022

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Updated translation

## Checklist:
- [ ] I have updated the changelog file accordingly, if necessary.

### DIFF
--- a/mRemoteNGTests/UI/Window/ConfigWindowTests/ConfigWindowGeneralTests.cs
+++ b/mRemoteNGTests/UI/Window/ConfigWindowTests/ConfigWindowGeneralTests.cs
@@ -288,7 +288,7 @@ namespace mRemoteNGTests.UI.Window.ConfigWindowTests
                         nameof(ConnectionInfo.SSHOptions),
                         nameof(ConnectionInfo.PuttySession),
                         nameof(ConnectionInfo.OpeningCommand),
-                        nameof(ConnectionInfo.UserViaAPI)
+                        nameof(ConnectionInfo.UserViaAPI),
 			nameof(ConnectionInfo.EC2InstanceId),
                     	nameof(ConnectionInfo.EC2Region)
                     });
--- a/mRemoteNGTests/UI/Window/ConfigWindowTests/ConfigWindowGeneralTests.cs
+++ b/mRemoteNGTests/UI/Window/ConfigWindowTests/ConfigWindowGeneralTests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -261,9 +261,10 @@ namespace mRemoteNGTests.UI.Window.ConfigWindowTests
 			nameof(ConnectionInfo.RdpVersion),
                         nameof(ConnectionInfo.RDPStartProgram),
                         nameof(ConnectionInfo.RDPStartProgramWorkDir),
+			nameof(ConnectionInfo.UserViaAPI)
                         nameof(ConnectionInfo.EC2InstanceId),
                         nameof(ConnectionInfo.EC2Region),
-                        nameof(ConnectionInfo.UserViaAPI)
+
                     });
                     break;
                 case ProtocolType.VNC:
@@ -274,6 +275,9 @@ namespace mRemoteNGTests.UI.Window.ConfigWindowTests
                         nameof(ConnectionInfo.Port),
                         nameof(ConnectionInfo.VNCSmartSizeMode),
                         nameof(ConnectionInfo.VNCViewOnly),
+			nameof(ConnectionInfo.UserViaAPI)
+			nameof(ConnectionInfo.EC2InstanceId),
+                    	nameof(ConnectionInfo.EC2Region),
                     });
                     break;
                 case ProtocolType.SSH1:
@@ -286,6 +290,8 @@ namespace mRemoteNGTests.UI.Window.ConfigWindowTests
                         nameof(ConnectionInfo.PuttySession),
                         nameof(ConnectionInfo.OpeningCommand),
                         nameof(ConnectionInfo.UserViaAPI)
+			nameof(ConnectionInfo.EC2InstanceId),
+                    	nameof(ConnectionInfo.EC2Region),
                     });
                     break;
                 case ProtocolType.SSH2:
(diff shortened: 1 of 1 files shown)


Vest:
@tecxx, please verify my PR just in case, because I fixed the tests that were broken after the new functionality was added.

Kvarkas:
Thanks for such, I am only worry if http/https should had change username to userviaapi as this fields seems specific only for amazon stuff...

Vest:
Can you please tell me how this feature (with user API) should work? I have simply debugged failed tests and tried to correct it. 
By the way, are your tests green, or some of them are red?

Kvarkas:
> Can you please tell me how this feature (with user API) should work? I have simply debugged failed tests and tried to correct it.
> By the way, are your tests green, or some of them are red?

TBH I am not sure, but can imaging on the call its filled via script then connect to amazon instance, its was some time ago added functionality but no much details provided and no update for documentation so lets wait if @tecxx can shed some light on the mater. 

Currently I have 21 test failing due stream is null.  (i do similar changes but was not yet push as it still red, so you jump behind me :D but its ok, technically its not make me worry to much as its only failure on test itself not impacting how mRemoteNG works... but not very sure :) so better to fix them of course.

tecxx:
hello,
first, sorry for breaking tests, i will happily help to fix them where possible. also let me know where / how to update the docs, i will do so as well.

there are currently two features in the external connectors dll.

#1 automatic update of host ip from amazon AWS EC2 instance id
fill the ec2 instance id and the amazon ec2 region identifier in the two corresponding fields
image

upon connection, the plugin will talk to amazon ec2 rest api and fetch the current public ip of this instance, and put it into hostname/ip field.
this way when the instance gets a new ip, you don't need to manually update it in mremote.
what is required to use this?
an ec2 access key, if this is not set in registry, a popup will show up and ask for it.


#2 plugin for retrieving credentials from thycotic secret server api
fill the secret id into the field "API User ID" and leave username, password and domain empty.
before opening a rdp or ssh connection, the plugin will access thycotic secret server and fetch your credentials from there. this way these sensitive secrets are never stored on your harddisk.
image

if you have not yet set up the connection to TSS, a popup will ask for connection information.



if you think that any of these fields should be renamed or replaced somewhere else, sure, just go ahead / let me know.
i'll try to look into the failed checks / your PR tonight.

tecxx:
"Every time when a new property is added to a connection node, the tests ConfigWindowGeneralTests should be corrected. The code there checks what properties must be available for different connection types, and what not."

if i understand this and your PR correctly (i'm not native english speaker), then i'd say

                        nameof(ConnectionInfo.EC2InstanceId),
                        nameof(ConnectionInfo.EC2Region),

should also be added to the SSH case (lines 288, 302) (feature #1)


and adding
                         nameof(ConnectionInfo.UserViaAPI),
to the https, powershell and IntApp cases is probably going to work, but i currently added the TSS connector (feature #2) only for RDP and SSH. so if there is need to also add this to these connection types, we can try and add it.

i added the changes to my local repo and this is how the tests currently look like:

image

Vest:
I don't know if it makes sense to mix a general connection with something related to Amazon. Maybe it deserves a separate group, or anything else. I am not the owner of this feature :)

The second point about SSH connection is that the tests showed that AWS EC2 specific fields (EC2InstanceId and EC2Region) exist for SSH2, but not for SSH1. And UserViaAPI exists almost everywhere, but not for VNC, I guess.

That's why I said that I don't know how this feature should work :)

@tecxx, if you use the current PR (without your modifications), are those tests "green"? If yes, this is what I told you - I don't know what connection types should support AWS, and what shouldn't.

Kvarkas:
@Vest Technically its sounds usable how ever I am not pointing to much attention in a form where and how its posted as planning to add profiles to user, connections and that will be slightly change. 

@tecxx its ok because all responsibility of broken test on me :D as I allow to merge yours PR, Test we are using is automated (a lot!) to help us to understand if we not miss something, however if they failing but not impact app work its kind of acceptable to me because functionality is more important (we are talking about dev version) btw what is yours native language?

I wish to make some changes as well so adding new option should be simple and easy - current setup not allow them even group properly - thinking to add support of local db instead of settings file so that's also will be a lot of changes (not functionality) 

Documentation is posted via GitHub pages (currently only in English) https://mremoteng.readthedocs.io/ you should have access to its repository: https://github.com/mRemoteNG/mRemoteNG/tree/develop/mRemoteNGDocumentation

For now I add mentioned changes and I will pull this request, so we can test tests (bla bla bla) :D however extremally busy at the moment with main work and family (thanks covid!) so may have some time to work more on such mater only on weekends.

Kvarkas:
@tecxx regarding "ec2 access key, if this is not set in registry, a popup will show up and ask for it." as I understand such key its kind of access token? do we need ability to save it to connection file (or db a bit later) or its dynamically changing?

tecxx:
trying to answer all questions :)
my native language is german.

the access key is some kind of token, yes. for a user, you can create access keys like in the screenshot below and then describe permissions for this key (e.g. readonly - it would be wise not use write permissions for this kind of access)
image

it is not dynamically changing, it only changes when the admin revokes the key and generates a new one for the user.

i have no good knowledge of mremote internal way of storing data so for me it was easiest to just show a popup and then store the values in a regkey. of course this can be moved to an internal storage like you said, and instead of showing popup a setting in the main config ui can be done. however as all this code is in a seperate DLL i am not sure if it's that easy to access those other parts of mremote there? maybe it's necessary to scratch the dll and move to code to main codebase?

"I don't know if it makes sense to mix a general connection with something related to Amazon. Maybe it deserves a separate group, or anything else. I am not the owner of this feature :)"

yes, i agree, and i think this can be rather easily extended for other cloud providers like google or microsoft azure. in that case it may make sense to add a dropdown field to the options list, like "external ip fetcher" with options "none, AWS, Azure, Google". and if "none" is selected, the additional fields don't even appear.
but i didn't yet take time to investigate how to add such dropdown or how to better group this in the UI :) 

"if you use the current PR (without your modifications), are those tests "green"?"
i posted a screenshot in the answer above with everything that shows as green, do you see it? SSH1, SSH2, RDP, all show up as green so i think it should be good! 

i'll see about adding info to the docs, thanks for link!

Kvarkas:
@tecxx thanks for such answer its become much clear now for me ;)
Summarising: one of my wish is to add category to the instance bit. So then adding new instance you user may select is that cloud instance (then depending on that clouds we supporting, I am Azure guy btw :)) we may add different fields, like in this case AWS seems don't need user password as it use that key - and technically we not interested to see IP as it will be updated (kind of grayout field) also interesting that will happened if user write ip and provide AWS region and instance id - should be overwritten I expect. But for now if its work and pass a tests its ok :)

Vest:
Thank you all for this discussion and result :)

PR #2123, opened 2022-01-14
Thycotic Secret Server -  Added 2FA OTP support

jacko873:
Added support for 2FA OTP
token management and use if a refresh token

## Description
Added ability to enter OTP token optionally, and adjusted authentication to reuse valid tokens and when possible use Refresh token to reduce the number of times users need to log in

## Motivation and Context
Secret Server is a password vault and its hard to imagine most users would not be using 2FA to secure their environment, This of course was the case for me. So i added it 

## How Has This Been Tested?
Tested against Thycotic cloud
With windows and Linux machines, 

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Updated translation

## Checklist:


Kvarkas:
@jacko873 could you please commit you changes for documentation https://github.com/mRemoteNG/mRemoteNG/tree/v1.77.3-dev/mRemoteNGDocumentation

tecxx:
@jacko873  initial connection to secret server worked, but having mremote open for a while now results in this error when trying subsequent connections:

----------
WarningMsg
01/17/2022 14:04:46
A connection protocol error occurred. Host: "xxxxx.local"; Error code: "0"; Error Description: "Secret Server Interface Error: One or more errors occurred. (Could not deserialize the response body stream as SecretServerAuthentication.TSS.TokenErrorResponse.

Status: 400
Response: 
)"
----------

didn't yet find the time to debug why this is happening.

jacko873:
> @jacko873 initial connection to secret server worked, but having mremote open for a while now results in this error when trying subsequent connections:
> 
> WarningMsg 01/17/2022 14:04:46 A connection protocol error occurred. Host: "xxxxx.local"; Error code: "0"; Error Description: "Secret Server Interface Error: One or more errors occurred. (Could not deserialize the response body stream as SecretServerAuthentication.TSS.TokenErrorResponse.
> 
> ## Status: 400
> Response:
> )"
> didn't yet find the time to debug why this is happening.

Hey, iv haven't seen this yet on my side, how long till it broke for you?

tecxx:
roughly an hour, i guess, which seems to correspond with my refresh token time (running it in debugger now to verify)

anyway i think something is missing in GetToken()

line 186 says "try using refresh token" - what if the code below can't use the refresh token? it throws an exception and fails. no popup is shown to update / reset credentials or OTP codes. 

the only place where a popup is shown is in function Init() and there is a check in the first line to just continue silently if password has already been set. how did you intend to re-enter a new OTP code once the old expired?

jacko873:
I think your right, I missed this, 

Before I go-ahead and adjust, I was thinking about adding the config to the main options > Credentials form.

one problem i had when i got the dev branch, so i could use your TSS code, was how to i set it up, the first place i went was option and credentials,

What are your thoughts on this?

tecxx:
TBH this is the first time i ever contributed to a project like this. i made the code already a year ago for internal use in my company. it was just a quick hack to "make it work" (and we use 2FA but in a different way, so we didn't need it in this particular place ;)

so honestly this can be changed to whatever is good for the project. just keep in mind this connector code is currently in a separate dll so i'm not sure how easy it is to have the credential form in the general options, and the connector code in the dll? and don't forget, since the code doesn't save the API password, each time before connecting to TSS we need some kind of popup to ask for credentials. it would be cumbersome to open the general options manually each time. 

regarding the issue itself, i'll check it in the debugger in an hour and see what's happening, because for us this needs to work without OTP as well - but we'll get there. cheers!

tecxx:
okay, reproduced it.
after waiting +- 1h30min, i triggered a new connection to TSS and traced the code

line 138 - bearer is not null, so jumps into IF
line 180 - expires on is now smaller than UtcNow, so goes to else branch
line 187++ fetches a new token with the existing refresh token
everything is fine, secret is fetched.

triggering ANOTHER connection to TSS now produced this exception

image

in line 190

image


i assume that either the expiration date and/or the bearer and refresh_token variables need to be updated after the first successful refresh token fetch?

PR #2188, opened 2022-03-18
Update session display settings

### LINKED ISSUE #1546: RDP Dynamic Resolution Support
Please add support for the [dynamic resolution](https://techcommunity.microsoft.com/t5/Enterprise-Mobility-Security/Resolution-and-Scaling-Level-Updates-in-RDP-8-1/ba-p/248077) feature that was introduced in RDP 8.1. This feature changes the server-side resolution to match the client window on resize without reconnecting. The effect is a clear, sharp desktop without scrollbars when resizing or switching from full screen to windowed mode. 

## Expected Behavior
A user should be able to start a RDP session at a set resolution, then change the window size and have the RDP session updated to match the new window dimensions without losing image quality. Switching from full-screen should resize the session to my full screen resolution automatically. Switching back to window should resize the session back to the original resolution.

You can see this feature in action using the Windows Store version of Remote Desktop. I believe it is also how Enhanced Mode works for Hyper-V, only this generalizes it to all session hosts.

## Current Behavior
The Smart Sizing feature keeps the server resolution the same as when connected and then scales the image up or down on resize. Scaling makes the screen unreadable by distorting the text and images or making it blurry.

To get close to dynamic resolution, you have to disconnect from the server and reconnect at the desired size. Workarounds have been put in place to automatically do this for users, but the experience is very slow because of how long it takes RDP to reestablish a connection as documented in https://github.com/mRemoteNG/mRemoteNG/issues/1152.

## Possible Solution
This requires adding support for the Dynamic Resolution changes introduced in 8.1.

Some references to teams that are looking to or have implemented this:
https://github.com/neutrinolabs/xrdp/issues/448
[update-screen-size-on-tablet-rotation-using-axmstsclib](https://social.technet.microsoft.com/Forums/en-US/94f51499-01b9-4a6a-9a34-b9dd95852a2e/update-screen-size-on-tablet-rotation-using-axmstsclib)

From what I gathered, the [IMsRdpClient9 interface](https://docs.microsoft.com/en-us/windows/win32/termserv/imsrdpclient9) needs to be used with the methods:
[SyncSessionDisplaySettings](https://docs.microsoft.com/en-us/windows/win32/termserv/imsrdpclient9-syncsessiondisplaysettings)
[UpdateSessionDisplaySettings](https://msdn.microsoft.com/en-us/library/Mt703457(v=VS.85).aspx)

It looks like it needs to have a Dynamic Virtual Channel set 
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpedisp/2878b488-794e-42c1-8f98-a23fa950def8
https://docs.microsoft.com/en-us/windows/win32/termserv/dvc-plug-in-registration
https://docs.microsoft.com/en-us/windows/win32/termserv/imstscadvancedsettings-plugindlls

## Context
Resizing has been a poor experience in RDP for a long time. Workarounds were added but none of them really made the experience better because the root problem was the server-side resolution. This is resolved by the Dynamic Resolution feature and just needs to be implemented in mRemoteNG.

Trellmor:
## Description

Use IMsRdpClient9::UpdateSessionDisplaySettings to dynamically update the
session display settings without reconnecting. RDP Version needs to be
Rdc9 or Highest for this to work.

## Motivation and Context

Fixes #1546

## How Has This Been Tested?

Tested locally by connecting and verifying that resolution changes properly.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Updated translation

This modifies the resize behaviour.

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 - #2138: Improve compatibility with Remote Desktop Connection Manager v2.83
 - #2123: Thycotic Secret Server - Added 2FA OTP support
+### Changed
+- #1546: Enable resize without reconnect for RDP Version Rdc9 or higher
  
 ## [1.77.2]
 ### Added
--- a/mRemoteNG/Connection/Protocol/RDP/RdpProtocol8.cs
+++ b/mRemoteNG/Connection/Protocol/RDP/RdpProtocol8.cs
@@ -29,7 +29,7 @@ namespace mRemoteNG.Connection.Protocol.RDP
             protected set
             {
                 base.SmartSize = value;
-                ReconnectForResize();
+                DoResizeClient();
             }
         }
 
@@ -39,7 +39,7 @@ namespace mRemoteNG.Connection.Protocol.RDP
             protected set
             {
                 base.Fullscreen = value;
-                ReconnectForResize();
+                DoResizeClient();
             }
         }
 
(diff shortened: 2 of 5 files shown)


Kvarkas:
Thanks for such

tecxx:
thanks a lot for resize feature. love it!

MrChriZ:
Should this work when the connection is undocked?  It seems to work well here when docked but undocked the resolution doesn't seem to change?

PR #2285, opened 2022-09-07
support extraction of SSH private keys from external cred prov

tecxx:
mremote can now extract ssh private keys from external credential provider (DSS) for ssh authentication

supported formats: rsa, rsa with passphrase, putty private key

### DIFF
--- a/ExternalConnectors/DSS/SecretServerInterface.cs
+++ b/ExternalConnectors/DSS/SecretServerInterface.cs
@@ -1,6 +1,11 @@
 ﻿using Microsoft.Win32;
+using Org.BouncyCastle.Crypto;
+using Org.BouncyCastle.Crypto.Parameters;
+using Org.BouncyCastle.OpenSsl;
+using Org.BouncyCastle.Security;
 using SecretServerAuthentication.DSS;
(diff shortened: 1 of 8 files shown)


Tampa:
rsa is on the way out, so other formats like ed25519 and ecdsa should be added as well.

Using pageant to provide keys can overlap with sessions and vice versa, so this becomes a breaking change for anyone using a local keychain along with a modified putty session. Changelog should definitely mention this.

tecxx:
i'm not sure i can follow, if you configure a connection to fetch credentials & key from the external vault, why would that interfere with local keychains? this PR does not modify mremote's default behaviour at all. only when you connect to DSS to specifically fetch a key, that key is passed to putty via -i switch. but in that case you want your keys stored in DSS anyway and not locally handled with pageant. correct me if i'm wrong.

Tampa:
I have two credential providers setup so when I compiled from source today to fix a small annoyance I had my putty session load from them rather than the set key in the session. It's an easy change, but depending on the existing setup might cause a bit of a stumble when it suddenly does something it was setup to do, but never did. Just something to mention on the changelog for whenever this hits release, all I'm saying. :)

PR #2931, opened 2025-10-17
Feature/vault openbao connector

Pernotto:
## Description
I implemented an initial version of a Vault/Openbao connector.

## Motivation and Context
[Hashicorp Vault](https://www.hashicorp.com/en/products/vault) and its fork [Openbao](https://openbao.org) have become increasingly useful and powerful tools for a secure environment.

## How Has This Been Tested?
I tried with my own Openbao cluster, my use cases are:
- RDP with credentials on KV engine
- RDP with static-cred on LDAP engine
- SSH with credentials on KV engine
- SSH with OTP on SSH engine (not working)

## Screenshots:
Node connection section for an Openbao/Vault managed LDAP secret
node

## Types of changes

- [x] Changed feature (non-breaking change which changes functionality)
- [ ] Updated translation

## Checklist:

- [ ] All Tests within VisualStudio are passing
- [ ] I have updated the changelog file accordingly, if necessary.

## TODO:
- Tried to run tests, but 97 did not passed. They were not from the code I added, maybe I'm doing something wrong.
- Many features of Openbao/Vault have not been implemented:
  - Openbao/Vault provides lots of types of credentials and certificates, it's impossible to implement all the ways they can be used.
  - Putty does not provide a way to pass Keyboard-interactive credentials, this prevents SSH-OTP to work
- Changelog & Documentation: I'm not sure if this approach is the best for using such a complex system
- There is no dev environment setup script. `vault server -dev` or `bao server -dev` are very useful but still needs a script to put a secret for every supported type.

## Known issues:
- The KV is the simplest and also the most abused secret engine. I assumed to use the "Username" as accessor to the key.
- It misses a complete error management for every use case

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -98,6 +98,7 @@
     <PackageVersion Include="System.ValueTuple" Version="4.6.1" />
     <PackageVersion Include="System.Windows.Extensions" Version="9.0.9" />
     <PackageVersion Include="System.Xml.ReaderWriter" Version="4.3.1" />
+    <PackageVersion Include="VaultSharp" Version="1.17.5.1" />
     <PackageVersion Include="VncSharpCore" Version="1.2.1" />
     <PackageVersion Include="ZstdSharp.Port" Version="0.8.6" />
   </ItemGroup>
--- a/ExternalConnectors/ExternalConnectors.csproj
+++ b/ExternalConnectors/ExternalConnectors.csproj
@@ -20,6 +20,7 @@
     <PackageReference Include="AWSSDK.EC2" />
     <PackageReference Include="BouncyCastle.Cryptography" />
     <PackageReference Include="Newtonsoft.Json" />
+    <PackageReference Include="VaultSharp" />
   </ItemGroup>
   <ItemGroup>
     <Compile Update="AWS\AWSConnectionForm.cs" />
--- a/ExternalConnectors/ExternalConnectors.csproj
+++ b/ExternalConnectors/ExternalConnectors.csproj
@@ -26,5 +26,8 @@
     <Compile Update="AWS\AWSConnectionForm.cs" />
     <Compile Update="CPS\CPSConnectionForm.cs" />
     <Compile Update="DSS\SSConnectionForm.cs" />
+    <Compile Update="VO\VaultOpenbaoConnectionForm.cs">
+      <SubType>Form</SubType>
+    </Compile>
   </ItemGroup>
 </Project>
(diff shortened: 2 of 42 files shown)


Kvarkas:
dont worry about tests, they are a bit broken :/

also keep in mind what in new version we dont store passwords in connection tree, but just ids and source from where it will be pooled.

Pernotto:
Thank you! I'll work on the exceptions handling next week.

Kvarkas:
also if you have time think how this could be done as extention - for now code is not ready and not in this release, but perhaps to a next one - I am planing to implement plugin system, so we could add plugins to work with a varity of the systems (and it they not required users could remove them or dont install) so we could have a bit cleaner structure and not a switch knifth what scary users with its multy complexity :)

Kvarkas:
in regards "to pass Keyboard-interactive credentials" we have a method what past into putty a command to execute after start, could be this used to pass any information you need to be pass?

Pernotto:
> also if you have time think how this could be done as extention - for now code is not ready and not in this release, but perhaps to a next one - I am planing to implement plugin system, so we could add plugins to work with a varity of the systems (and it they not required users could remove them or dont install) so we could have a bit cleaner structure and not a switch knifth what scary users with its multy complexity :)

That would wonderful! I wonder how can you achieve the UI integration for connectors that need extra dialogs, do you plan to move from winforms?

Pernotto:
> in regards "to pass Keyboard-interactive credentials" we have a method what past into putty a command to execute after start, could be this used to pass any information you need to be pass?

Keyboard-interactive is a request from the server, user must answer to access. It is not the same as password even if the prompt may ask for a password. 
I found an example putty plugin: https://git.tartarus.org/?p=simon/putty.git;a=blob;f=contrib/authplugin-example.py;hb=HEAD
Assuming everybody using openbao/vault use [vault-ssh-helper](https://github.com/hashicorp/vault-ssh-helper) to enable OTP on the server it is possible to handle that specific case.
Maybe plink?

Pernotto:
I developed an authentication plugin for PuTTY: https://github.com/MaxPlap/putty-vault-ssh-helper
I tested it and it works, allowing me to authenticate to vault-ssh-helper secured server without manually copy-paste the OTP.
mRemoteNG should communicate with it using a named pipe to pass the credentials.
Is there a way to pass the plugin as a parameter to PuttyNG?

Kvarkas:
Hello, need to check but should work, maybe need some changes in what we sending into putty but on first look is doable

PR #3122, opened 2026-02-08
Add-master-password-autolock-on-minimize-idle-1649

### LINKED ISSUE #1649: When a master password is defined, mRemoteNG may be lock after a while or when minimized
## Expected Behavior

First of all, thanks for your great software ! 

(this feature is present in KeePass, a password manager)

1) Add an option under "configuration" (root of the connexion tree), near "Password protect". Only if password protect is set to "yes", then this option can be enabled. This option is : "Autolock" (yes/no). The help message could be something like "Autolock after 5 minutes of user inactivity, or when minimized."

2) When this option is set to yes, and after 5 minutes without user interaction inside mRemoteNG (or maybe inside Windows, but a global hook will be necessary in that case, https://github.com/gmamaladze/globalmousekeyhook may help in both cases), or when mRemoteNG is minimized, then mRemoteNG lock itself : blank panel, with the master dialog box opened (to unlock).

Maybe these additional configs may be useful :
- seconds before lock [ number of seconds ]
- lock when minimized [ yes / no ]
- lock by pressing [ a key ]

## Context

The reason of this suggestion : recently, a pc in our company was hacked, and the hacker destroy 3 of our servers (thanks Lord, we have crypted backups ! so no damage), because mRemoteNG was still open on a pc (our fault), easy for him to connect to these servers. I think it's hard to be auto-disciplined to close everytime mRemoteNG, so an autolock function would be great ! The Windows auto-lock feature may seem enough, but developers don't want Windows to lock every 5mn. So I think this feature will be a good security improvement.

robertpopa22:
## Summary 
- Add root-level `Auto lock on minimize` option, visible only when password protection is enabled. 
- Persist `AutoLockOnMinimize` in root XML attributes. 
- Introduce runtime autolock behavior: 
  - lock when app is minimized (if option enabled) 
  - lock after 5 minutes of system inactivity (if option enabled) 
- Require master password to unlock when restoring from tray/taskbar. 
- Add regression tests for root default value and XML serialize/deserialize of the new flag. 
 
## Issue 
- Closes #1649

### DIFF
--- a/mRemoteNG/App/NativeMethods.cs
+++ b/mRemoteNG/App/NativeMethods.cs
@@ -104,6 +104,9 @@ namespace mRemoteNG.App
         [DllImport("user32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
         internal static extern int GetClassName(IntPtr hWnd, StringBuilder lpClassName, int nMaxCount);
 
+        [DllImport("user32.dll", SetLastError = true)]
+        internal static extern bool GetLastInputInfo(ref LASTINPUTINFO plii);
+
         [DllImport("user32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
         internal static extern int GetDlgCtrlID(IntPtr hwndCtl);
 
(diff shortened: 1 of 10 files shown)


robertpopa22:
CI verification for this patch package: 
 
- Fork validation run: CI run 21794326986 on `robertpopa22/mRemoteNG` 
- Result: success on solution builds (x86/x64/ARM64) and tests/specs build jobs (x86/x64).

---

> *We sincerely appreciate the mRemoteNG maintainers and the wonderful community that has kept this project alive and thriving for over a decade. Your work powers thousands of IT professionals worldwide, including us. Our fork builds entirely on your foundation — we're preparing a consolidated Pull Request to contribute these improvements back to the official project. Thank you for everything you've built.*
>
> *— robertpopa22, community contributor*

PR #3379, opened 2026-07-13
fix: Quick Connect resolution, Spanish Sessions translation, and panel order

Die2FF:
## Summary

Fixes three separate issues:

### 1. Quick Connect resolution / screen sizing
Connections opened from the Quick Connect toolbar rendered blurry and undersized (tiny icons/text) because they used the default `SmartSize` resolution, which connects at full-screen resolution and scales the image down to the panel. Quick Connect connections have no parent, so they cannot inherit the sharper `FitToWindow` resolution that tree/panel connections get from their container. They are now forced to `FitToWindow` so they render sharp and sized to the window, matching connections opened from the panel.

### 2. Missing Spanish translation
The `_Sessions` key was missing from `Language.es.resx`, so the main toolbar fell back to English ("Sessions" instead of "Sesiones"). Added the translation.

### 3. Connections/Config panel order
The Connections panel is now reliably ordered above the Config panel, including when they are separate auto-hide panes on the same edge. `EnforcePanelOrder` was made reusable and is now also re-applied by `PanelBinder` after toggling auto-hide, so the order stays correct at runtime.

## Testing
Verified manually on Windows:
- Quick Connect RDP sessions now render sharp and correctly sized, matching panel-opened connections.
- Main toolbar shows "Sesiones" in Spanish.
- Connections panel stays above Config panel across restarts and auto-hide toggles.
- Solution builds successfully (Release/x64).

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,6 @@
+{
+    "recommendations": [
+        "ms-dotnettools.csharp",
+        "ms-dotnettools.csdevkit"
+    ]
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,43 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Launch mRemoteNG (Debug x64)",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "build",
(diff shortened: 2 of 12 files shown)


Die2FF:
Closing: changes are intended to stay in my personal fork only.

PR #3406, opened 2026-08-05
fix: apply the colour picked in the property grid Color/TabColor fields

jafin:
## Problem

Setting **Color** or **Tab Color** on a connection from the property grid's colour picker fails with `Property value is not valid` / `Object of type 'System.Drawing.Color' cannot be converted to type 'System.String'`, and the colour is never applied.

## Root cause

Both properties are `string`, but they are annotated with the stock `System.Drawing.Design.ColorEditor`, which returns a `System.Drawing.Color`. The PropertyGrid commits an editor's return value straight through `PropertyDescriptor.SetValue` and never runs the `TypeConverter` on that path, so the assignment throws.

`MiscTools.TabColorConverter` already handles `Color` → `string`, but only the text-entry path goes through it — which is why it does not prevent this error.

## Changes

- **`ColorStringEditor`** (new) — wraps `ColorEditor` for string-backed colour properties: feeds the editor a `Color`, translates the picked colour back into the stored string form. It also restores the swatch preview, which is blank today because `ColorEditor` cannot paint a string value.
- `AbstractConnectionRecord.Color` / `.TabColor` now use the new editor.
- `TabColorConverter.GetStandardValues` returns strings instead of `Color` objects — picking an entry from the dropdown list fails in exactly the same way.
- `TabColorConverter.ConvertFrom(Color.Empty)` now yields `""` instead of `"#00000000"`, so clearing a colour leaves the property unset.

## Verification

`mRemoteNG.csproj` builds clean with the change (MSBuild, `Release|x64`).

Behaviour was verified by exercising the built `mRemoteNG.dll` from a standalone harness — all checks pass:

- string round-trip through the editor for `"Red"`, `"#804020"`, `""` and `null`
- both properties resolve to `ColorStringEditor`
- `SetValue` with the editor's output stores `"Purple"` instead of throwing
- standard values are all strings

New tests: `mRemoteNGTests/UI/Controls/ColorStringEditorTests.cs` (non-interactive — no `IWindowsFormsEditorService` is supplied, so no dialog is shown) plus an empty-colour case in `TabColorConverterTests`.

### Note on the test project

`mRemoteNGTests` does not currently compile on `v1.78.2-dev`, independently of this PR:

```
mRemoteNGTests/Config/Settings/OptionsRepositoryTests.cs(25,49): error CS1503:
Argument 1: cannot convert from 'mRemoteNG.Config.Settings.Store.OptionsStore'
to 'mRemoteNG.Config.Settings.Store.ISettingsStore'
```

`OptionsStore` implements `IDisposable` but not `ISettingsStore`, while `OptionsRepository` takes an `ISettingsStore`. That break predates this branch and is unrelated to colour handling, so I have left it alone rather than widen the scope here. With that one pre-existing file excluded, the test project — including the new tests — compiles cleanly.

### DIFF
--- a/mRemoteNG/Connection/AbstractConnectionRecord.cs
+++ b/mRemoteNG/Connection/AbstractConnectionRecord.cs
@@ -165,7 +165,7 @@ namespace mRemoteNG.Connection
         [LocalizedAttributes.LocalizedCategory(nameof(Language.Display)),
          LocalizedAttributes.LocalizedDisplayName(nameof(Language.Color)),
          LocalizedAttributes.LocalizedDescription(nameof(Language.PropertyDescriptionColor)),
-         Editor(typeof(System.Drawing.Design.ColorEditor), typeof(System.Drawing.Design.UITypeEditor)),
+         Editor(typeof(UI.Controls.ConnectionInfoPropertyGrid.ColorStringEditor), typeof(System.Drawing.Design.UITypeEditor)),
          TypeConverter(typeof(MiscTools.TabColorConverter))]
         public virtual string Color
         {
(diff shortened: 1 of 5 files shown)


jafin:
Follow-up pushed: `TabColorConverter` maps text it cannot parse to `Color.Empty` rather than throwing, so a legacy or hand-edited value such as `not-a-color` reached the wrapped `ColorEditor` as an empty colour. Dismissing the picker echoed that empty colour back and it was stored as an empty string, silently discarding the original value.

`ColorStringEditor` now tracks whether the stored value actually describes a colour (`TryConvertToColor`) and leaves the value untouched when it does not and no colour was picked. A genuinely unset value — `null` or blank — still round-trips to an empty string, and picking a real colour still stores its string form.

Covered by new cases in `ColorStringEditorTests.EditValueKeepsAValueItCannotParse` (`not-a-color`, `#nothex`).

PR #3408, opened 2026-08-06
fix: keep the active tab when a tab close is cancelled

jafin:
## Problem

Closing a connection tab shows the "are you sure you want to close" prompt. Clicking **Cancel** correctly keeps the tab open, but the active connection switches to the tab on its left.

## Root cause

Not in the confirmation logic — it is DockPanelSuite. `DockPaneStripBase.TryCloseTab`:

```csharp
DockPane.CloseContent(content);
if (PatchController.EnableSelectClosestOnClose == true)
    SelectClosestPane(index);        // DockPane.ActiveContent = Tabs[index - 1].Content
```

`SelectClosestPane` runs unconditionally — it never checks whether the close actually happened. `EnableSelectClosestOnClose` defaults to `true`, and `connDock.DocumentStyle` is `DockingWindow` whenever more than one tab is open, so both of its guards pass. When `ConnectionTab.OnFormClosing` sets `e.Cancel = true`, the tab survives but DPS has already moved the selection.

It only bites when the tab being closed is not the first one (`SelectClosestPane` is guarded by `index > 0`), which is why the bug can look intermittent.

## Fix

`DockPaneStripNG.TabCloseButtonHit` now goes through a wrapper that puts the selection back when the tab is still on display:

```csharp
private void CloseTab(int index)
{
    IDockContent tabContent = Tabs[index].Content;
    IDockContent? activeBeforeClose = DockPane.ActiveContent;

    TryCloseTab(index);

    if (activeBeforeClose == null || ReferenceEquals(DockPane.ActiveContent, activeBeforeClose))
        return;

    if (DockPane.DisplayingContents.Contains(tabContent) &&
        DockPane.DisplayingContents.Contains(activeBeforeClose))
    {
        activeBeforeClose.DockHandler.Activate();
    }
}
```

Keying off `DisplayingContents` rather than a cancel flag covers every refusal — the confirmation prompt, a protocol veto — and it deliberately does nothing on a successful close, so the existing selection behaviour is untouched. `Activate()` rather than a bare `ActiveContent` assignment also restores keyboard focus to the tab.

## Verification

`mRemoteNG.csproj` builds clean (MSBuild, `Release|x64`).

Behaviour was verified by driving the real `DockPaneStripNG` from a standalone STA harness against the built assembly:

```
PASS  Doc2 starts active
PASS  cancelled close leaves Doc2 open
PASS  cancelled close keeps Doc2 active
  (control) after raw TryCloseTab, Doc2 open=True, active=Doc1
PASS  uncancelled close still closes Doc3
```

The control line is the bug reproduced directly on this branch: calling DPS's `TryCloseTab` leaves the tab open but hands the selection to `Doc1`.

New tests in `mRemoteNGTests/UI/Tabs/DockPaneStripNGTests.cs` cover both the cancelled close and a normal close. They are headless — the cancel is simulated with a `FormClosing` handler, so no dialog is shown — and run on an STA thread with a message pump.

### Note on the test project

`mRemoteNGTests` does not currently compile on `v1.78.2-dev`, independently of this PR:

```
mRemoteNGTests/Config/Settings/OptionsRepositoryTests.cs(25,49): error CS1503:
Argument 1: cannot convert from 'mRemoteNG.Config.Settings.Store.OptionsStore'
to 'mRemoteNG.Config.Settings.Store.ISettingsStore'
```

`OptionsStore` implements `IDisposable` but not `ISettingsStore`, while `OptionsRepository` takes an `ISettingsStore`. That break predates this branch and is unrelated, so I have left it alone rather than widen the scope. With that one pre-existing file excluded, the test project — including the new tests — compiles cleanly.

### DIFF
--- a/mRemoteNG/UI/Tabs/DockPaneStripNG.cs
+++ b/mRemoteNG/UI/Tabs/DockPaneStripNG.cs
@@ -1183,7 +1183,35 @@ namespace mRemoteNG.UI.Tabs
             Point mousePos = PointToClient(MousePosition);
             Rectangle tabRect = GetTabBounds(Tabs[index]);
             if (tabRect.Contains(ActiveClose) && ActiveCloseHitTest(mousePos))
-                TryCloseTab(index);
+                CloseTab(index);
+        }
+
+        /// <summary>
+        /// Closes the tab at <paramref name="index"/>, keeping the current selection when
+        /// the tab does not actually close.
+        /// </summary>
+        /// <remarks>
+        /// DockPanelSuite's TryCloseTab selects the neighbouring tab afterwards without
+        /// checking whether the close happened, so dismissing the "are you sure" prompt
+        /// still switched the active connection.
+        /// </remarks>
+        private void CloseTab(int index)
+        {
+            IDockContent tabContent = Tabs[index].Content;
+            IDockContent? activeBeforeClose = DockPane.ActiveContent;
+
+            TryCloseTab(index);
+
+            if (activeBeforeClose == null || ReferenceEquals(DockPane.ActiveContent, activeBeforeClose))
+                return;
+
+            // The tab is still on display, so the close was refused and the selection
+            // moved for nothing.
+            if (DockPane.DisplayingContents.Contains(tabContent) &&
+                DockPane.DisplayingContents.Contains(activeBeforeClose))
+            {
+                activeBeforeClose.DockHandler.Activate();
+            }
         }
 
         private Rectangle GetCloseButtonRect(Rectangle rectTab)
--- a/mRemoteNGTests/UI/Tabs/DockPaneStripNGTests.cs
+++ b/mRemoteNGTests/UI/Tabs/DockPaneStripNGTests.cs
@@ -1,0 +1,158 @@
+using System;
+using System.Reflection;
+using System.Threading;
+using System.Windows.Forms;
+using mRemoteNG.Themes;
+using NUnit.Framework;
+using WeifenLuo.WinFormsUI.Docking;
+using WeifenLuo.WinFormsUI.ThemeVS2015;
+
+namespace mRemoteNGTests.UI.Tabs
+{
+    [TestFixture]
+    public class DockPaneStripNGTests
+    {
+        private static void RunWithMessagePump(Action testAction)
+        {
+            Exception? caught = null;
+            var thread = new Thread(() =>
+            {
+                try
+                {
+                    testAction();
+                }
+                catch (Exception ex)
+                {
+                    caught = ex;
+                }
+            });
+
+            thread.SetApartmentState(ApartmentState.STA);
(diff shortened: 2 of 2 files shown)


jafin:
Review feedback addressed:

- The second test now asserts `dockPaneStrip` and `closeTabMethod` are non-null before use, matching the first test, so a missing control or renamed method fails with a clear message instead of an `NullReferenceException`.
- Dropped the `Exception?` annotation that raised CS8632. I went with option 2 rather than adding `<Nullable>enable</Nullable>` to `mRemoteNGTests.csproj` — that would change how every existing test file in the project compiles and belongs in its own change. After this, the only CS8632 warnings left in the test project are the pre-existing ones in `Tools/Registry/WindowsRegistryTests.cs`.

Both tests pass:

```
Passed CancellingATabClose_LeavesTheActiveTabUnchanged [900 ms]
Passed ClosingATabThatIsNotCancelled_StillClosesIt [256 ms]
```

(Run with `SupportedOSPlatformVersion` overridden at build time to 10.0.17763.0 — the committed value of 10.0.26100.0 stops NUnit running on my Windows 10.0.22631 host. No project file was changed.)

jafin:
Reworked `RunWithMessagePump` as suggested:

- The test body is posted with `BeginInvoke` onto an STA thread running `Application.Run(new ApplicationContext())`, and calls `Application.ExitThread()` in a `finally`, so the loop unwinds as soon as the body finishes (or throws).
- On timeout the loop is asked to exit via `BeginInvoke(Application.ExitThread)` and joined again for 5s before `Assert.Fail`.
- `IsBackground = true` is the last-resort net: a thread that never pumps again still cannot hold the run open. `Thread.Interrupt()` is gone — it only unblocks interruptible waits and gave a false sense of cancellation.

Verified the helper's three paths with a standalone copy using a 2s timeout, since exercising the real 30s path in a test would just be slow:

```
PASS  normal body runs to completion
PASS  body exception propagates
PASS  wedged body times out
PASS  timeout returns promptly (< 15s)
  timeout path took 7.1s          (2s timeout + 5s shutdown join)
```

The wedged body was `Thread.Sleep(Timeout.Infinite)` with no pumping; the process still exited immediately after the failure was reported, which is the behaviour that was missing before.

Both tests pass against a clean rebuild:

```
Passed CancellingATabClose_LeavesTheActiveTabUnchanged [848 ms]
Passed ClosingATabThatIsNotCancelled_StillClosesIt [249 ms]
```

One thing worth noting for anyone copying this helper: `BeginInvoke(new MethodInvoker(...))` does not compile in this file, because `System.Reflection` is imported and `MethodInvoker` is ambiguous with `System.Reflection.MethodInvoker` on .NET 8+. It uses `Action` instead. Similarly the catch is `InvalidOperationException` only — `ObjectDisposedException` derives from it, so a separate clause is unreachable (CS0160).

PR #3411, opened 2026-08-06
security: fix command injection in Terminal protocol (#3335)

### LINKED ISSUE #3335: Malicious .xml executes any command when triggering a connection via the file or automatically
mRemoteNG's Terminal protocol constructs a `cmd.exe /K` command by directly concatenating unsanitized connection fields into an argument string. Opening a malicious connections file triggers arbitrary OS command execution under the victim's user context.

## Affected Code

`mRemoteNG/Connection/Protocol/Terminal/Connection.Protocol.Terminal.cs`, lines 53–74:

```csharp
string sshCommand = "ssh";
if (!string.IsNullOrEmpty(username))
    sshCommand += $" {username}@{_connectionInfo.Hostname}";  // no sanitization
else
    sshCommand += $" {_connectionInfo.Hostname}";             // no sanitization

arguments = $"/K {sshCommand}";
_consoleControl.StartProcess(terminalExe, arguments);         // cmd.exe /K <injected>
```

`terminalExe` resolves to `%COMSPEC%` (typically `cmd.exe`). The `&`, `|`, `>`, `<`, `^` metacharacters are unblocked. `Hostname` and `Username` are read directly from the connections XML with no validation at any layer.

---

## Exploitation

An attacker delivers a crafted `.xml` connections file. The victim opens the file and double-clicks the connection entry.

On installations where **Reconnect to previously opened sessions** is enabled (`OpenConsFromLastSession = true`, `OptionsStartupExitPage.settings`), no click is required — the connection fires automatically on file load via `PreviousSessionOpener`.

**Hostname payload:**
```
a & cmd /c echo pwned > C:\Users\Public\poc_vuln1.txt & rem
```

Encoded in XML:
```xml
Hostname="a &amp; cmd /c echo pwned &gt; C:\Users\Public\poc_vuln1.txt &amp; rem "
```

Resulting invocation:
```
cmd.exe /K ssh a & cmd /c echo pwned > C:\Users\Public\poc_vuln1.txt & rem
```

`ssh a` fails immediately; the injected command executes unconditionally via `&`.

**Trigger behavior:**
- ConsoleControl launches `cmd.exe` with `CreateNoWindow = true` — no visible console
- `Terminal.cs` then deadlocks on the UI thread for 5 seconds waiting for a WinForms handle that can never be created (unparented control, blocked message pump)
- mRemoteNG logs "Connection failed" after the timeout
- The subprocess has already executed by this point — the failure is irrelevant

---

## Impact

Arbitrary command execution as the logged-in user. No elevation required. mRemoteNG is typically deployed by system administrators managing privileged infrastructure — the user context is high-value. The connection entry name is attacker-controlled and can appear legitimate within a shared connections file.

jafin:
Fixes #3335

## Problem

The Terminal protocol built `cmd.exe /K ssh <user>@<host>` by concatenating the **unsanitized** `Hostname`/`Username` connection fields into a single argument string that `cmd.exe` then re-parsed. Because `%COMSPEC%` is a shell, its metacharacters (`&  |  <  >  ^`) are interpreted, so a malicious `.xml` connections file achieves arbitrary command execution as the logged-in user.

Example `Hostname`:

```
a & cmd /c echo pwned > C:\Users\Public\poc.txt & rem
```

produces:

```
cmd.exe /K ssh a & cmd /c echo pwned > C:\Users\Public\poc.txt & rem
```

`ssh a` fails; the injected command runs unconditionally via `&`. With **Reconnect to previously opened sessions** enabled (`OpenConsFromLastSession`), `PreviousSessionOpener` fires the connection automatically on file load — zero clicks. The console is launched with `CreateNoWindow = true`, so nothing is visible.

I reproduced this locally against the exact argument construction: the injected command executed and wrote the PoC file.

## Fix

Launch **`ssh.exe` directly** instead of routing through `cmd.exe`, so no shell interprets the arguments.

- `BuildSshArguments` validates `Hostname` and `Username` and rejects any value that contains whitespace/control characters or starts with `-` (which `ssh` would treat as an option — e.g. `-oProxyCommand=...` argument injection). It builds a properly tokenized `[-p <port>] [user@]host` argument.
- `FindSshExe` locates the Windows OpenSSH client (`%SystemRoot%\System32\OpenSSH\ssh.exe`, then `PATH`); if absent, the connection fails with an actionable message instead of falling back to a shell.
- Local sessions (empty/`localhost` hostname) still open `%COMSPEC%` with a fixed `/K` and **no** user-controlled input.

This mirrors the existing `ProtocolOpenSSH` approach of invoking `ssh.exe` directly.

## Tests

Adds `ProtocolTerminalTests` (21 cases) covering:

- **Valid targets:** `host`, `user@host`, non-default port (`-p 2222 ...`), default/zero port omission, whitespace trimming, IPv6 (`[fe80::1]`).
- **Rejected:** the issue's `&` payload, shell metacharacters (`| & > < ^`), embedded whitespace/tab/newline/CR, and leading-`-` option injection in both hostname and username.

All 21 pass. The main `mRemoteNG` project compiles cleanly against `v1.78.2-dev`.

> Note: the test project on `v1.78.2-dev` has a **pre-existing, unrelated** compile error in `OptionsRepositoryTests.cs` (`OptionsRepository(OptionsStore)` vs `ISettingsStore`) that predates this change; I left it untouched. The new tests were verified green on an equivalent base.

### DIFF
--- a/mRemoteNG/Connection/Protocol/Terminal/Connection.Protocol.Terminal.cs
+++ b/mRemoteNG/Connection/Protocol/Terminal/Connection.Protocol.Terminal.cs
@@ -37,8 +37,8 @@ namespace mRemoteNG.Connection.Protocol.Terminal
                     Padding = new Padding(0, 20, 0, 0)
                 };
 
-                string hostname = _connectionInfo.Hostname.Trim().ToLower();
-                bool useLocalHost = hostname == "" || hostname.Equals("localhost");
+                string hostname = _connectionInfo.Hostname.Trim();
+                bool useLocalHost = hostname.Length == 0 || hostname.Equals("localhost", StringComparison.OrdinalIgnoreCase);
 
                 string processExe;
                 string arguments;
@@ -152,11 +152,14 @@ namespace mRemoteNG.Connection.Protocol.Terminal
             string hostname = (rawHostname ?? string.Empty).Trim();
             string username = (rawUsername ?? string.Empty).Trim();
 
+            // Deliberately do NOT echo the offending value: it is attacker-controlled and, in the
+            // invalid case, may contain control characters/newlines that would be written verbatim into
+            // logs and UI messages (log forging / message spoofing).
             if (!IsSafeSshToken(hostname))
-                throw new ArgumentException($"Refusing to start SSH session: the hostname '{hostname}' contains characters that are not allowed.");
+                throw new ArgumentException("Refusing to start SSH session: the hostname contains characters that are not allowed.");
 
             if (username.Length > 0 && !IsSafeSshToken(username))
-                throw new ArgumentException($"Refusing to start SSH session: the username '{username}' contains characters that are not allowed.");
+                throw new ArgumentException("Refusing to start SSH session: the username contains characters that are not allowed.");
 
             string args = "";
 
@@ -173,8 +176,11 @@ namespace mRemoteNG.Connection.Protocol.Terminal
 
         /// <summary>
         /// Returns true only for values that are safe to place on the ssh.exe command line as a single
-        /// token: non-empty, no whitespace or control characters, and not starting with '-' (which ssh
-        /// would treat as an option switch).
+        /// token: non-empty, no whitespace or control characters, no double quotes, and not starting
+        /// with '-' (which ssh would treat as an option switch). Double quotes are rejected because
+        /// Windows argument parsing (CommandLineToArgvW) strips them, so a value such as
+        /// "-oProxyCommand=..." does not literally start with '-' here yet reaches ssh.exe as an argv
+        /// token that does — re-enabling option injection.
         /// </summary>
         private static bool IsSafeSshToken(string value)
         {
@@ -186,7 +192,7 @@ namespace mRemoteNG.Connection.Protocol.Terminal
 
             foreach (char c in value)
             {
-                if (char.IsWhiteSpace(c) || char.IsControl(c))
+                if (char.IsWhiteSpace(c) || char.IsControl(c) || c == '"')
                     return false;
             }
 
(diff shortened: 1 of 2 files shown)


jafin:
## Follow-up: review hardening (commit b731546)

Addressed four review findings on the `ssh.exe` launch path. All changes are in `Connection.Protocol.Terminal.cs`, plus regression tests.

**1. Culture-sensitive localhost check**
Replaced `Hostname.Trim().ToLower()` + `== ""` with `Length == 0` and `Equals("localhost", StringComparison.OrdinalIgnoreCase)` — avoids the Turkish-I class of bugs and a throwaway allocation.

**2. Quote-based option injection in `IsSafeSshToken`**
The validator blocked a leading `-`, but not quotes. Windows `CommandLineToArgvW` strips quotes, so a token like `"-oProxyCommand=calc.exe"` passes the leading-`-` check yet reaches `ssh.exe` as an argv token that *does* start with `-`. `"` is now rejected anywhere in hostname/username. IPv6 bracket form (`[fe80::1]`) is unaffected.

**3. `FindSshExe` PATH parsing edge cases**
Each PATH segment is now normalized before probing: strip surrounding quotes, skip empty segments, expand `%VAR%` placeholders, and skip non-rooted entries — so we never build a quoted or relative `ssh.exe` candidate.

**4. Log/UI injection via exception messages**
`BuildSshArguments` validation errors no longer echo the offending hostname/username. That value is attacker-controlled and, in the invalid case, may contain control characters/newlines that would be written verbatim into logs and UI messages (log forging / message spoofing). Messages are now generic, so no change to `MessageCollector`/`MiscTools` was needed.

**Tests**
Added regression coverage for quoted-leading-dash and embedded-quote payloads (both hostname and username) and for the sanitized error message.

**Verification note**
The NUnit suite can't execute in my environment: this branch targets `net10.0-windows10.0.26100.0` but my host is Windows build 22631, so NUnit refuses all Windows-targeted tests (the branch also has a pre-existing, unrelated `OptionsRepositoryTests` compile break). I verified the fixes against the compiled assembly via reflection instead — 14/14 checks pass (quoted/embedded-quote rejection, sanitized single-line messages, all prior protections, and valid targets including IPv6). The main project builds clean. CI on a `windows-2025` runner will exercise the full NUnit suite.

PR #3412, opened 2026-08-07
feat(port scan): single address field (IP / range / CIDR, IPv4 + IPv6) and a port mode selector

jafin:
## What

The Port Scan panel's two input areas collapse into one control each.

**Scan target** — one field replaces the *First IP* / *Last IP* octet boxes and accepts, in IPv4 **or IPv6** form:

| Notation | Example |
|---|---|
| single address | `192.168.1.1`, `2001:db8::1` |
| explicit range | `192.168.1.1 - 192.168.1.254` (spaces optional, endpoints in either order) |
| CIDR block | `192.168.1.0/24`, `2001:db8::/120` |

**Ports** — a three-way choice replaces the *First Port* / *Last Port* checkboxes and their spinners:

```
Ports   ( ) Common ports   ( ) All ports   (o) Custom
        [ 22, 80, 443, 3389, 8000-8100                    ]
```

*Common ports* (the default) probes 34 well-known service ports — including every port behind a protocol column in the results list, so the SSH/Telnet/HTTP/HTTPS/Rlogin/RDP/VNC columns are still populated. *All ports* is 1–65535. *Custom* takes a list of ports and ranges, deduplicated and sorted; the field is only editable while *Custom* is selected, so the port options cannot be left half-configured.

## Why

The old panel could only express an IPv4 range, and only as two four-octet controls — there was no way to scan a single host, a CIDR block, or anything IPv6 at all. The port side had two independent checkboxes that silently fell back to "default ports only" when neither was ticked.

## How

- `Tools/IpRangeParser` and `Tools/PortListParser` are new, self-contained and unit tested; the window just calls them. A failed parse now reports *why* (bad address, mixed IPv4/IPv6 endpoints, prefix out of range, invalid port) instead of the generic `CannotStartPortScan` warning.
- `MrngIpTextBox` becomes a single-line text box that takes a whole address, replacing the four-octet `UserControl` that could only express IPv4. `PortScanWindow` was its only consumer.
- `PortScanner` gains a constructor taking an explicit port list, and enumerates address ranges as **unsigned** `BigInteger`s instead of `Int32`. That fixes ordering across the whole space (an IPv4 range straddling `128.0.0.0` used to invert, because the octets were read as a signed integer) and lets an IPv6 range be enumerated at all. A 65,536-address cap keeps an IPv6 range — which can span an astronomically large number of addresses — from exhausting memory; exceeding it is reported to the user rather than throwing into the log.
- The scanning engine itself (ping + `TcpClient` per port on a background thread) is unchanged.

New user-facing strings are added to `Language.resx` / `Language.Designer.cs` (neutral English) — `PortScanAddressRange`, `PortScanAddressRangeHint`, `PortScanCommonPorts`, `PortScanAllPorts`, `PortScanCustomPorts`, `PortScanCustomPortsHint`. `FirstIp`, `LastIp`, `FirstPort` and `LastPort` are left in place rather than removed from all 30 translations.

Second commit is a drive-by on the same panel: the Open Ports / Closed Ports columns rendered `22, 80, 443, ` — the trailing separator is gone.

## Verification

- `mRemoteNG` builds clean on `v1.78.2-dev` with these changes.
- The 42 new tests in `IpRangeParserTests` / `PortListParserTests` pass.
- Two caveats on this branch's base, both pre-existing and untouched here: `mRemoteNGTests` does not compile at `v1.78.2-dev` (`OptionsRepositoryTests` passes an `OptionsStore` where an `ISettingsStore` is expected), and the test project's `SupportedOSPlatformVersion` of `10.0.26100.0` means NUnit skips every test on a host below Windows 11 24H2. I set both aside locally to run the suite; neither is changed by this PR. With them set aside, the only other failures in the `Tools` group are `ExternalToolTests` (argument passing) and `TabColorConverterTests` (named vs ARGB colour equality), which are likewise pre-existing and unrelated.


jafin:
Both localization comments are addressed in 524d37e3f.

**Placeholders.** Added `PortScanAddressRangePlaceholder` and `PortScanCustomPortsPlaceholder` to `Language.resx` / `Language.Designer.cs`; `ApplyLanguage()` reads both from `Language` instead of assigning literals.

**Validation and reporting messages.** I went with suggestion #2 (templates in `Language.resx`, referenced from the tool classes) rather than the structured error codes of #1. `mRemoteNG.Tools` already depends on `mRemoteNG.Resources.Language` in nine files — `ScanHost`, `MiscTools`, `ExternalTool`, `SecureTransfer`, `NotificationAreaIcon` among them — so an error-code enum plus a UI-side mapping table would introduce a layering convention the project doesn't otherwise follow, for the same user-visible result. Happy to switch to #1 if you'd rather establish that boundary here.

Five new templates:

| Key | Used by |
|---|---|
| `PortScanInvalidAddress` | `IpRangeParser.TryParseAddress` |
| `PortScanInvalidPrefixLength` | `IpRangeParser.TryParseCidr` |
| `PortScanMixedAddressFamilies` | `IpRangeParser.TryParseRange` and `PortScanner.IpAddressArrayFromRange` |
| `PortScanInvalidPort` | `PortListParser.TryParsePort` |
| `PortScanRangeTooLarge` | the `MaxScanRange` cap in `PortScanner` |

That covers the oversized-range guard, which `StartScan()` surfaces through `ReportInvalidInput(ex.Message)`.

Both `SyntaxHint` members now resolve to the existing `PortScanAddressRangeHint` / `PortScanCustomPortsHint` resources instead of duplicating their text. That changes them from `const` to `static` properties, since resource lookups aren't compile-time constants; no caller depended on const-ness. Formatting uses `string.Format(CultureInfo.CurrentCulture, ...)`, matching the `string.Format(Language.X, ...)` pattern already used across `App/`.

**Verification.** Release/x64 rebuild of `mRemoteNG.csproj` is clean with no new warnings on the changed lines. `IpRangeParserTests` + `PortListParserTests`: 42 passed, 0 failed — they assert the error is non-empty and contains the offending token, both preserved by the templates.

Kvarkas:
@copilot review suggestion and implement into plugin ip scaner

